PR #1, opened 2023-04-27
suggested changes

trel:
for the `no-trash` option of `data-objects?op=remove`... perhaps...  consider `use-trash` with `default: 1` instead?


korydraughn (approved):

korydraughn:
I'll write up issues for things like `no-trash` vs `use-trash`.

I'm also preparing to make the API.md file into a Wiki.

Issue #40, opened 2023-06-29
Update use of C++ client connection libraries to match changes in 4-3-stable


korydraughn:
See irods/irods#7156 for changes to libraries.

PR #41, opened 2023-06-29
[#40] Updated use of C++ client connection libraries

korydraughn:


korydraughn:
Ignoring clang-tidy and clang-format for this PR.

Will handle them in a separate/later PR.

Adding pounds.

korydraughn:
Added #'s

korydraughn:
@alanking Ping.

Not sure if you receive notifications for this repo.

src/main.cpp
```diff
@@ -210,10 +210,8 @@ auto init_irods_connection_pool(const json& _config) -> irods::connection_pool
         conn_pool.at("size").get<int>(),
         client.at("host").get_ref<const std::string&>(),
         client.at("port").get<int>(),
-        username,
-        zone,
-        username,
-        zone,
+        {{username, zone}},
```
alanking, line 213: Why double curly braces?

alanking (commented):
My bad, probably got caught in the swirl...

Just had one question. Everything else looks fine

src/main.cpp
korydraughn: std::optional<fully_qualified_username>

The first set of braces are for the optional. The second set are for the inner type.

alanking: Oh okay, cool. Thanks!

alanking (approved):

Issue #55, opened 2023-07-12
Add support for TLS between the HTTP API and iRODS


korydraughn:

korydraughn:
Initial investigation indicates the HTTP API can't support this through its configuration file because the iRODS SSL plugin reads settings from the environment. The plugin does not allow settings to be passed to it.

That means SSL can only be supported by either providing a local .irods/irods_environment.json file or defining environment variables that map to the appropriate members of the `rodsEnv` structure.

Using environment variables isn't horrible, but doesn't allow the HTTP API to be as polished as we'd like.

korydraughn:
Perhaps we can get around this by having the HTTP API server set the environment properties itself on startup.

If the properties are defined in the HTTP API config file, then make the necessary calls to set them in the environment. The call to `getRodsEnv` would pick them up and no one has to worry about managing the environment variables themselves.

iRODS does this today using the `environment_variables` property in server_config.json.

trel:
Having the HTTP API setting environment variables for itself seems fine/normal.   Bet this will work.  Will need to make sure we provide a 'clean' environment, so we don't pollute anything else or accidentally inherit things we don't want (but, also... perhaps inherit things we 'do' want?).

korydraughn:
Very true.

I suspect making the HTTP API set the env vars is safe, assuming it is working in an entirely new environment (because fork-exec / sub-shell / new process).

Changing env vars should not leak into the parent process's space.

trel:
agreed

korydraughn:
If all of this is true (and can be proved), then the only thing needed to support TLS/SSL is the defining the structure of the TLS stanza, figuring out which properties are required, and naming those properties.

PR #70, opened 2023-08-02
[#55] Added support for TLS/SSL between HTTP API and iRODS

korydraughn:
Verified at the bench that everything works.

Just need to figure out how to write a test for it. This may require more machinery since setting up an iRODS server is out-of-scope for this.

I also need to update the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,46 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g. 
         // The zone of the target iRODS server.
         "zone": "<string>",
 
+        // Defines options for secure communication with the target iRODS server.
+        "tls": {
+            // Controls whether the client and server communicate using TLS.
+            //
+            // The following values are supported:
+            // - CS_NEG_REFUSE:    Do not use secure communication.
+            // - CS_NEG_REQUIRE:   Demand secure communication.
+            // - CS_NEG_DONT_CARE: Let the server decide.
+            "client_server_policy": "CS_NEG_REFUSE",
+
+            // The file containing trusted CA certificates in PEM format.
+            //
+            // Note that the certificates in this file are used in conjunction
+            // with the system default trusted certificates.
+            "ca_certificate_file": "<string>",
+
+            // The file containing the server's certificate chain.
+            //
+            // The certificates must be in PEM format and must be sorted
+            // starting with the subject's certificate (actual client or server
+            // certificate), followed by intermediate CA certificates if
+            // applicable, and ending at the highest level (root) CA.
+            "certificate_chain_file": "<string>",
+
+            // The file containing Diffie-Hellman parameters.
+            "dh_params_file": "<string>",
+
+            // Defines the level of server certificate authentication to
+            // perform.
+            //
+            // The following values are supported:
+            // - none:     Authentication is skipped.
+            // - cert:     The server verifies the certificate is signed by
+            //             a trusted CA.
+            // - hostname: Equivalent to "cert", but also verifies the FQDN
+            //             of the iRODS server matches either the common
+            //             name or one of the subjectAltNames.
+            "verify_server": "cert"
+        },
+
         // The credentials for the rodsadmin user that will act as a proxy
         // for all authenticated users.
         "proxy_rodsadmin": {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6,6 +6,7 @@
 #include "version.hpp"
 
 #include <irods/connection_pool.hpp>
+#include <irods/irods_configuration_keywords.hpp>
 #include <irods/process_stash.hpp>
 #include <irods/rcConnect.h>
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -130,6 +130,47 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
         // The zone of the target iRODS server.
         "zone": "<string>",
 
+        // Defines options for secure communication with the target iRODS server.
+        "ssl": {
```
korydraughn, line 134: Perhaps this should be `tls`?

Which is more correct? SSL or TLS?

trel: TLS is more correct/updated.

trel: "Transport Layer Security (TLS) is the upgraded version of SSL that fixes existing SSL vulnerabilities."

trel: We need to go through and update the iRODS server as well one day, but that's a bigger lift.

Out here, we should use the good/new words.

korydraughn: Updated to use `tls`.

README.md
```diff
@@ -130,6 +130,14 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
         // The zone of the target iRODS server.
         "zone": "<string>",
 
+        "ssl": {
+            "client_server_policy": "CS_NEG_REFUSE",
+            "ca_certificate_file": "/path/to/file",
+            "certificate_chain_file": "/path/to/file",
```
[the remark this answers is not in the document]
trel: The `ca_certificate_file` is required.

The `certificate_chain_file` could also be sufficient, as its first entry is usually the same as the entry in the `certificate_chain_file`... but we use it as supplemental information to the `certificate_chain_file`.

src/main.cpp
```diff
@@ -190,6 +192,14 @@ auto print_configuration_template() -> void
         "port": 1247,
         "zone": "<string>",
 
+        "tls": {{
+            "client_server_policy": "<string>",
+            "ca_certificate_file": "<string>",
+            "certificate_chain_file": "<string>",
```
trel, line 198: alpha?   not sure it matters.

korydraughn: Will reorder.

And no, order doesn't matter here.

korydraughn:
I'm going to capture the testing of TLS support in a new issue.

I'm confident this works.

README.md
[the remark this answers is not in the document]
korydraughn: That has been my experience with testing this at the bench.

I've provided documentation that almost mirrors the words from docs.irods.org so I think we're good here.

src/main.cpp
korydraughn: I reordered the options, but the felt it was a little weird since some of them are for files and others control the behavior.

Perhaps grouping them like so is better:
```json
{
    "client_server_policy": "CS_NEG_REFUSE",
    "verify_server": "cert",

    "ca_certificate_file": "/path/to/file",
    "certificate_chain_file": "/path/to/file",
    "dh_params_file": "/path/to/file"
}
```

In all honesty, the order that is defined in the PR right now is fine to me.

Perhaps `client_server_policy` should be renamed to something like `negotiation_policy`?

I think `negotiation_policy` is better, but are there other phrases that express _This client requires X for TLS_?

Maybe `secure_communication_policy`?

Thoughts on all of this?

korydraughn:
Created #71 to discuss testing TLS/SSL.

src/main.cpp
trel: Ordering is fine as is - documentation is what matters.

---

`tls_policy` or `negotiation_policy`... maybe... but...

Let's not change the names - these names match the iRODS docs... unless you feel they are misleading in some way.

korydraughn: No, the more I look at it, the more I feel it is fine.

If `client_server_policy` was outside the `tls` stanza, that would make the property very confusing. It being part of the `tls` stanza makes it a bit more clear in what it is trying to convey.

NFSRODS doesn't use `client_server_policy` at all. Instead, it uses `ssl_negotiation_policy`.

Changing it to `client_server_negotiation_policy` is very clear to me.

trel: so you're suggesting changing one attribute from

`client_server_policy` to `client_server_negotiation_policy`?

korydraughn: Hmm. Well, I'm trying to remove confusion. IDK if what I just proposed makes since either.

Are those property names clear to you in regard to what we're trying to express?

To me, the property is meant to convey ... _For TLS, this app will behave as XYZ._

korydraughn: Ultimately, I'm fine with the current name.

The documentation helps and it sharing a name very close to what's in the irods_environment.json file isn't a bad thing either.

Can always tweak later if we want.

korydraughn: I'm going to leave things as they are.

We can discuss this on the road to 1.0.

PR #75, opened 2023-08-15
Implement Authorization Code grant OIDC flow

MartinFlores751:
Implemented a new OIDC flow. Rebased onto main and resolved several issues of rebasing thereafter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,8 @@ cmake_minimum_required(VERSION 3.18.0 FATAL_ERROR)
 
 find_package(IRODS REQUIRED CONFIG)
 
+include(RequireOutOfSourceBuild)
+
 set(IRODS_HTTP_API_BINARY_NAME "irods_http_api")
 
 set(IRODS_CLIENT_VERSION "0.1.0")
@@ -13,9 +15,10 @@ set(IRODS_HTTP_API_VERSION "${IRODS_CLIENT_VERSION}")
 
 set(IRODS_PACKAGE_REVISION "1")
 
-include(RequireOutOfSourceBuild)
-include(IrodsCXXCompiler)
+include(IrodsExternals)
+IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH(JWTCPP jwt-cpp0.6.99.0-0)
 
+include(IrodsCXXCompiler)
 set(CMAKE_CXX_STANDARD ${IRODS_CXX_STANDARD})
 set(CMAKE_CXX_EXTENSIONS OFF)
 # export-dynamic so stacktrace entries from executables have function names.
@@ -52,6 +55,10 @@ set(IRODS_BUILD_WITH_WERROR ${IRODS_BUILD_WITH_WERROR_DEFAULT} CACHE BOOL "Choos
 if (IRODS_BUILD_WITH_WERROR)
   add_compile_options(-Werror)
 endif()
+
+# jwt library to handle OIDC JWT responses containing user information
+find_package(jwt-cpp REQUIRED
+  HINTS "${IRODS_EXTERNALS_FULLPATH_JWTCPP}")
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fpermissive>)
--- a/README.md
+++ b/README.md
@@ -93,6 +93,24 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g. 
                 // The amount of time before a user's "Basic" authentication
                 // token expires.
                 "timeout_in_seconds": 3600
+            },
+
+            // Defines required OIDC related configuration.
+            "oidc": {
+                // The hostname or ip address of the OIDC provider.
+                "config_host": "oidc.example.org",
+
+                // The port the OIDC provider is listening on.
+                "port": 8080,
+
+                // Well known endpoint to retrieve configuration of OIDC provider.
+                "well_known_uri": "/realms/example",
+
+                // The client id given to the application by OIDC provider.
+                "client_id": "irods_http_api",
+
+                // URI pointing to the irods HTTP API auth endpoint.
+                "redirect_uri": "https://<domain>/irods-http/0.1.0/authenticate"
             }
         },
 
--- a/src/endpoints/authentication/CMakeLists.txt
+++ b/src/endpoints/authentication/CMakeLists.txt
@@ -26,4 +26,5 @@ target_link_libraries(
   ${IRODS_HTTP_ENDPOINT}
   PRIVATE
   CURL::libcurl
+  jwt-cpp::jwt-cpp
   nlohmann_json::nlohmann_json)
--- a/src/endpoints/authentication/impl.cpp
+++ b/src/endpoints/authentication/impl.cpp
@@ -19,125 +19,532 @@
 #include <boost/beast.hpp>
 #include <boost/beast/http.hpp>
 
+#include <iterator>
 #include <nlohmann/json.hpp>
 
+#include <jwt-cpp/jwt.h>
+#include <jwt-cpp/traits/nlohmann-json/traits.h>
+
+#include <fmt/core.h>
+
+#include <curl/curl.h>
+#include <curl/urlapi.h>
+
+#include <algorithm>
 #include <array>
 #include <chrono>
 #include <string>
 #include <string_view>
+#include <unordered_map>
 #include <vector>
+
+// clang-format off
+namespace beast = boost::beast; // from <boost/beast.hpp>
+namespace net   = boost::asio;  // from <boost/asio.hpp>
+// clang-format on
+
+using body_arguments = std::unordered_map<std::string, std::string>;
 
 namespace irods::http::handler
 {
-    // NOLINTNEXTLINE(performance-unnecessary-value-param)
-    IRODS_HTTP_API_ENDPOINT_ENTRY_FUNCTION_SIGNATURE(authentication)
-    {
-        if (_req.method() != boost::beast::http::verb::post) {
-            return _sess_ptr->send(fail(status_type::method_not_allowed));
-        }
+	auto hit_token_endpoint(std::string _encoded_body) -> nlohmann::json
+	{
+		const auto token_endpoint{
+			irods::http::globals::oidc_endpoint_configuration().at("token_endpoint").get_ref<const std::string&>()};
 
-        irods::http::globals::background_task([fn = __func__, _sess_ptr, _req = std::move(_req)] {
-            const auto& hdrs = _req.base();
-            const auto iter = hdrs.find("authorization");
-            if (iter == std::end(hdrs)) {
-                return _sess_ptr->send(fail(status_type::bad_request));
-            }
+		// Setup net
+		net::io_context io_ctx;
+		net::ip::tcp::resolver tcp_res{io_ctx};
+		beast::tcp_stream tcp_stream{io_ctx};
 
-            log::debug("{}: Authorization value: [{}]", fn, iter->value());
+		// Setup curl
+		CURLU* endpoint{curl_url()};
+		if (endpoint == nullptr) {
+			std::abort();
+		}
 
-            //
-            // TODO Here is where we determine what form of authentication to perform (e.g. Basic or OIDC).
-            //
+		// Parse url
+		CURLUcode rc{curl_url_set(endpoint, CURLUPART_URL, token_endpoint.c_str(), 0)};
+		if (rc != 0) {
+			log::debug("Something happend....");
+		}
 
-            const auto pos = iter->value().find("Basic ");
-            if (std::string_view::npos == pos) {
-                return _sess_ptr->send(fail(status_type::bad_request));
-            }
+		// Get host
+		char* host{};
+		rc = curl_url_get(endpoint, CURLUPART_HOST, &host, 0);
+		if (rc != 0) {
+			log::debug("Something happend....");
+		}
 
-            constexpr auto basic_auth_scheme_prefix_size = 6;
-            std::string authorization{iter->value().substr(pos + basic_auth_scheme_prefix_size)};
-            boost::trim(authorization);
-            log::debug("{}: Authorization value (trimmed): [{}]", fn, authorization);
+		// Get service/port
+		const auto port{std::to_string(irods::http::globals::oidc_configuration().at("port").get<int>())};
 
-            constexpr auto max_creds_size = 128;
-            std::uint64_t size = max_creds_size;
-            std::array<std::uint8_t, max_creds_size> creds{};
-            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-            const auto ec = irods::base64_decode(reinterpret_cast<unsigned char*>(authorization.data()), authorization.size(), creds.data(), &size);
-            log::debug("{}: base64 - error code=[{}], decoded size=[{}]", fn, ec, size);
+		// Get path
+		char* path{};
+		rc = curl_url_get(endpoint, CURLUPART_PATH, &path, 0);
+		if (rc != 0) {
+			log::debug("Something happend....");
+		}
 
-            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
-            std::string_view sv{reinterpret_cast<char*>(creds.data()), size}; 
-            //log::debug("{}: base64 decode credentials = [{}]", fn, sv); // BE CAREFUL!!! THIS LOGS THE USER'S PASSWORD!
+		// Addr
+		const auto resolve{tcp_res.resolve(host, port)};
 
-            const auto colon = sv.find(':');
-            if (colon == std::string_view::npos) {
-                return _sess_ptr->send(fail(status_type::unauthorized));
-            }
+		// TCP thing
+		tcp_stream.connect(resolve);
 
-            std::string username{sv.substr(0, colon)};
-            std::string password{sv.substr(colon + 1)};
-            //log::debug("{}: username=[{}], password=[{}]", fn, username, password); // BE CAREFUL!!! THIS LOGS THE USER'S PASSWORD!
+		// Build Request
+		constexpr auto version_number{11};
+		beast::http::request<beast::http::string_body> req{beast::http::verb::post, path, version_number};
+		req.set(beast::http::field::host, host);
+		req.set(beast::http::field::user_agent, irods::http::version::server_name);
+		req.set(beast::http::field::content_type,
+		        "application/x-www-form-urlencoded"); // Possibly set a diff way?
 
-            bool login_successful = false;
+		// Send
+		req.body() = std::move(_encoded_body);
+		req.prepare_payload();
 
-            try {
-                static const auto& rodsadmin_username = irods::http::globals::configuration().at(nlohmann::json::json_pointer{"/irods_client/proxy_admin_account/username"}).get_ref<const std::string&>();
-                static const auto& rodsadmin_password = irods::http::globals::configuration().at(nlohmann::json::json_pointer{"/irods_client/proxy_admin_account/password"}).get_ref<const std::string&>();
-                static const auto& zone = irods::http::globals::configuration().at(nlohmann::json::json_pointer{"/irods_client/zone"}).get_ref<const std::string&>();
+		// Send request
+		beast::http::write(tcp_stream, req);
 
-                CheckAuthCredentialsInput input{};
-                username.copy(input.username, sizeof(CheckAuthCredentialsInput::username));
-                zone.copy(input.zone, sizeof(CheckAuthCredentialsInput::zone));
+		// Read back req
+		beast::flat_buffer buffer;
+		beast::http::response<beast::http::string_body> res;
+		beast::http::read(tcp_stream, buffer, res);
 
-                namespace adm = irods::experimental::administration;
-                const adm::user_password_property prop{password, rodsadmin_password};
-                const auto obfuscated_password = irods::experimental::administration::obfuscate_password(prop);
-                obfuscated_password.copy(input.password, sizeof(CheckAuthCredentialsInput::password));
+		log::debug("Got the following resp back: {}", res.body());
 
-                int* correct{};
+		// Close socket
+		beast::error_code ec;
+		tcp_stream.socket().shutdown(net::ip::tcp::socket::shutdown_both, ec);
 
-                // NOLINTNEXTLINE(cppcoreguidelines-owning-memory, cppcoreguidelines-no-malloc)
-                irods::at_scope_exit free_memory{[&correct] { std::free(correct); }};
+		// Free up all items created in reverse order
+		curl_free(path);
+		curl_free(host);
 
-                auto conn = irods::get_connection(rodsadmin_username);
+		// Done
+		curl_url_cleanup(endpoint);
 
-                if (const auto ec = rc_check_auth_credentials(static_cast<RcComm*>(conn), &input, &correct); ec < 0) {
-                    log::error("{}: Error verifying native authentication credentials for user [{}]: error code [{}].", fn, username, ec);
-                }
-                else {
-                    log::debug("{}: correct = [{}]", fn, fmt::ptr(correct));
-                    log::debug("{}: *correct = [{}]", fn, (correct ? *correct : -1));
-                    login_successful = (correct && 1 == *correct);
-                }
-            }
-            catch (const irods::exception& e) {
-                log::error("{}: Error verifying native authentication credentials for user [{}]: {}", fn, username, e.client_display_what());
-            }
-            catch (const std::exception& e) {
-                log::error("{}: Error verifying native authentication credentials for user [{}]: {}", fn, username, e.what());
-            }
+		// JSONize response
+		return nlohmann::json::parse(res.body());
+	}
 
-            if (!login_successful) {
-                return _sess_ptr->send(fail(status_type::unauthorized));
-            }
+	auto encode_body(const body_arguments& _args) -> std::string
+	{
+		auto encode_pair{[](const body_arguments::value_type& i) {
+			return fmt::format("{}={}", irods::http::encode(i.first), irods::http::encode(i.second));
+		}};
 
-            static const auto seconds = irods::http::globals::configuration().at(nlohmann::json::json_pointer{"/http_server/authentication/basic/timeout_in_seconds"}).get<int>();
-            auto bearer_token = irods::process_stash::insert(authenticated_client_info{
-                .auth_scheme = authorization_scheme::basic,
-                .username = std::move(username),
-                .password = std::move(password),
-                .expires_at = std::chrono::steady_clock::now() + std::chrono::seconds{seconds}
-            });
+		return std::transform_reduce(
+			std::next(std::cbegin(_args)),
+			std::cend(_args),
+			encode_pair(*std::cbegin(_args)),
+			[](const auto& a, const auto& b) { return fmt::format("{}&{}", a, b); },
+			encode_pair);
+	}
 
-            response_type res{status_type::ok, _req.version()};
-            res.set(field_type::server, irods::http::version::server_name);
-            res.set(field_type::content_type, "text/plain");
-            res.keep_alive(_req.keep_alive());
-            res.body() = std::move(bearer_token);
-            res.prepare_payload();
+	auto is_error_response(const nlohmann::json& _response_to_check) -> bool
+	{
+		if (const auto error{_response_to_check.find("error")}; error != std::cend(_response_to_check)) {
+			std::string token_error_log;
+			token_error_log.reserve(500);
 
-            return _sess_ptr->send(std::move(res));
-        });
-    } // authentication
-} // namespace irods::http::endpoint
+			auto error_log_itter{fmt::format_to(
+				std::back_inserter(token_error_log), "{}: Token request failed! Error: [{}]", __func__, *error)};
+
+			// Optional OAuth 2.0 error parameters follow
+			if (const auto error_description{_response_to_check.find("error_description")};
+			    error_description != std::cend(_response_to_check))
+			{
+				error_log_itter = fmt::format_to(error_log_itter, ", Error Description [{}]", *error_description);
+			}
+
+			if (const auto error_uri{_response_to_check.find("error_uri")}; error_uri != std::cend(_response_to_check))
+			{
+				error_log_itter = fmt::format_to(error_log_itter, ", Error URI [{}]", *error_uri);
+			}
+
+			log::warn(token_error_log);
+			return true;
+		}
+
+		return false;
+	}
+
+	auto decode_username_and_password(std::string_view _encoded_data) -> std::pair<std::string, std::string>
+	{
+		std::string authorization{_encoded_data};
+		boost::trim(authorization);
+		log::debug("{}: Authorization value (trimmed): [{}]", __func__, authorization);
+
+		constexpr auto max_creds_size = 128;
+		std::uint64_t size{max_creds_size};
+		std::array<std::uint8_t, max_creds_size> creds{};
+		// NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+		const auto ec = irods::base64_decode(
+			reinterpret_cast<unsigned char*>(authorization.data()), authorization.size(), creds.data(), &size);
+		log::debug("{}: base64 - error code=[{}], decoded size=[{}]", __func__, ec, size);
+
+		// NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+		std::string_view sv{reinterpret_cast<char*>(creds.data()), size};
+
+		const auto colon = sv.find(':');
+		if (colon == std::string_view::npos) {
+			return {"", ""};
+		}
+
+		std::string username{sv.substr(0, colon)};
+		std::string password{sv.substr(colon + 1)};
+
+		return {username, password};
+	}
+
+	IRODS_HTTP_API_ENDPOINT_ENTRY_FUNCTION_SIGNATURE(authentication)
+	{
+		if (_req.method() == boost::beast::http::verb::get) {
+			url url;
+			bool did_except{false};
+			try {
+				url = irods::http::parse_url(_req);
+			}
+			catch (irods::exception& e) {
+				did_except = true;
+			}
+
+			if (did_except) {
+				irods::http::globals::background_task([fn = __func__, _sess_ptr, _req = std::move(_req)] {
+					body_arguments args{
+						{"client_id",
+					     irods::http::globals::oidc_configuration().at("client_id").get_ref<const std::string&>()},
+						{"response_type", "code"},
+						{"scope", "openid"},
+						{"redirect_uri",
+					     irods::http::globals::oidc_configuration().at("redirect_uri").get_ref<const std::string&>()},
+						{"state", "placeholder"}};
+
+					const auto auth_endpoint{irods::http::globals::oidc_endpoint_configuration()
+					                             .at("authorization_endpoint")
+					                             .get_ref<const std::string&>()};
+					const auto encoded_url{fmt::format("{}?{}", auth_endpoint, encode_body(args))};
+
+					log::debug("{}: Proper redirect to [{}]", fn, encoded_url);
+
+					response_type res{status_type::found, _req.version()};
+					res.set(field_type::server, irods::http::version::server_name);
+					res.set(field_type::location, encoded_url);
+					res.keep_alive(_req.keep_alive());
+					res.prepare_payload();
+
+					return _sess_ptr->send(std::move(res));
+				});
+			}
+			else {
+				irods::http::globals::background_task([fn = __func__,
+				                                       _sess_ptr,
+				                                       _req = std::move(_req),
+				                                       url = std::move(url)] {
+					// Will always be in response, as we always send it out
+					const auto state_iter{url.query.find("state")};
+
+					// Invalid/Fake request... Should have state query param
+					if (state_iter == std::end(url.query)) {
+						log::warn(
+							"{}: Received an Authorization response with no 'state' query parameter. Ignoring.", fn);
+						return _sess_ptr->send(fail(status_type::bad_request));
+					}
+
+					auto is_state_valid{[](std::string_view in_state) { return in_state.compare("placeholder") == 0; }};
+
+					// The state is invalid (i.e. doesn't exist, or have been used)
+					if (!is_state_valid(state_iter->second)) {
+						log::warn(
+							"{}: Received an Authorization response with an invalid 'state' query parameter. Ignoring.",
+							fn);
+						return _sess_ptr->send(fail(status_type::bad_request));
+					}
+
+					// Will only be available if authorization was successful
+					const auto code_iter{url.query.find("code")};
+
+					// Code does not exist, process response for error details...
+					if (code_iter == std::end(url.query)) {
+						const auto error_iter{url.query.find("error")};
+
+						// Required error parameter missing, malformed response
+						if (error_iter == std::end(url.query)) {
+							log::warn(
+								"{}: Received an Authorization response with no 'code' or 'error' query parameters. "
+								"Ignoring.",
+								fn);
+							return _sess_ptr->send(fail(status_type::bad_request));
+						}
+						std::string responses;
+						responses.reserve(500);
+
+						auto responses_iter{fmt::format_to(
+							std::back_inserter(responses), "{}: Error Code [{}]", fn, error_iter->second)};
+
+						// Optional OAuth 2.0 error parameters follow
+						const auto error_description_iter{url.query.find("error_description")};
+						if (error_description_iter != std::end(url.query)) {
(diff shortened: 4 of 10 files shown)


korydraughn:
Will review as soon as I can.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+					const std::string jwt_token{res_item.at("id_token").get_ref<const std::string&>()};
+
+					// Get OIDC token && feed to JWT parser
+					// TODO: Handle case where we throw!!!
+					auto decoded_token{
+						jwt::decode<jwt::traits::nlohmann_json>(res_item.at("id_token").get_ref<const std::string&>())};
+
+					// Get irods username
+					const std::string irods_name{
+						decoded_token.get_payload_json().at("irods_username").get<const std::string>()};
```
korydraughn, line 259: Looks like this is where we have to think about user mappings, etc

Am I correct?

_Not in this PR though._

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+
+					// Issue token?
+					static const auto seconds =
+						irods::http::globals::configuration()
+							.at(nlohmann::json::json_pointer{"/http_server/authentication/basic/timeout_in_seconds"})
```
korydraughn, line 264: The OIDC config needs to grow its own timeout setting.

No shared state between basic and oidc please.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+							.get<int>();
+
+					auto bearer_token = irods::process_stash::insert(authenticated_client_info{
+						.auth_scheme = authorization_scheme::basic,
```
korydraughn, line 268: basic -> oidc

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+						.username = std::move(irods_name),
+						.expires_at = std::chrono::steady_clock::now() + std::chrono::seconds{seconds}});
+
+					response_type res_rep{status_type::ok, _req.version()};
+					res_rep.set(field_type::server, BOOST_BEAST_VERSION_STRING);
```
korydraughn, line 273: Replace macro

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+			}
+		}
+		// Handle posts
+		else if (_req.method() == boost::beast::http::verb::post) {
+			irods::http::globals::background_task([fn = __func__, _sess_ptr, _req = std::move(_req)] {
+				// Right, we're kinda being a proxy, so how about proxy-authorization?
+				const auto& hdrs{_req.base()};
+				const auto iter{hdrs.find("authorization")};
+
+				// Failed to find auth header, use OIDC auth_code flow instead
```
korydraughn, line 290: Incorrect comment?

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+
+					// BEGIN BASE64 HEADER DECODE
+
+					constexpr auto basic_auth_scheme_prefix_size = 6;
+					std::string authorization{iter->value().substr(alt_method + basic_auth_scheme_prefix_size)};
+					boost::trim(authorization);
+					log::debug("{}: Authorization value (trimmed): [{}]", fn, authorization);
+
+					constexpr auto max_creds_size = 128;
+					unsigned long size = max_creds_size;
```
korydraughn, line 322: Prefer a more explicit type.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+
+					const auto colon = sv.find(':');
+					if (colon == std::string_view::npos) {
+						return _sess_ptr->send(fail(status_type::unauthorized));
+					}
+
+					std::string username{sv.substr(0, colon)};
+					std::string password{sv.substr(colon + 1)};
+					log::debug(
+						"{}: username=[{}], password=[{}]", fn, username, password); // TODO Don't print the password
```
korydraughn, line 345: Remove password.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+					const std::string jwt_token{res_item.at("id_token").get<const std::string>()};
+
+					// Feed to JWT parser
+					auto decoded_token{jwt::decode<jwt::traits::nlohmann_json>(jwt_token)};
+
+					// Get irods username
+					// Zones?
+					// uname#zname
+					const std::string irods_name{
+						decoded_token.get_payload_json().at("irods_username").get<const std::string>()};
```
korydraughn, line 369: const reference?

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+
+					// Issue token?
+					static const auto seconds =
+						irods::http::globals::configuration()
+							.at(nlohmann::json::json_pointer{"/http_server/authentication/basic/timeout_in_seconds"})
```
korydraughn, line 374: basic -> oidc

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+							.get<int>();
+					auto bearer_token = irods::process_stash::insert(authenticated_client_info{
+						.auth_scheme = authorization_scheme::basic,
```
korydraughn, line 377: basic -> oidc

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+						.username = std::move(irods_name),
+						.expires_at = std::chrono::steady_clock::now() + std::chrono::seconds{seconds}});
+
+					response_type res_rep{status_type::ok, _req.version()};
+					res_rep.set(field_type::server, BOOST_BEAST_VERSION_STRING);
```
korydraughn, line 382: Replace macro.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+					return _sess_ptr->send(std::move(res_rep));
+				}
+
+				constexpr auto basic_auth_scheme_prefix_size = 6;
+				std::string authorization{iter->value().substr(pos + basic_auth_scheme_prefix_size)};
+				boost::trim(authorization);
+				log::debug("{}: Authorization value (trimmed): [{}]", fn, authorization);
+
+				constexpr auto max_creds_size = 128;
+				unsigned long size = max_creds_size;
```
korydraughn, line 397: Looks like things were lost/reverted in the rebase from this point on (this line included).

Please compare the original logic for basic auth to what you have in here and restore the lines that were changed.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+					nlohmann::json res_item{hit_token_endpoint(encode_body(args))};
+
+					// Assume passed, get oidc token
+					const std::string jwt_token{res_item.at("id_token").get<const std::string>()};
+
+					// Feed to JWT parser
+					auto decoded_token{jwt::decode<jwt::traits::nlohmann_json>(jwt_token)};
+
+					// Get irods username
+					// Zones?
```
korydraughn, line 366: No need to worry about zones. iRODS users can only authenticate against the zone which they were created in AFAIK.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+			encode_pair);
+	}
+
+	auto get_encoded_redirect_uri() -> std::string
+	{
+		return encode_string(
+			irods::http::globals::oidc_configuration().at("redirect_uri").get_ref<const std::string&>());
+	}
+
+	auto authentication(session_pointer_type _sess_ptr, request_type& _req) -> void
```
korydraughn, line 178: Lost the HANDLER macro in the rebase. See the original code.

korydraughn (commented):
Good job!

README.md
```diff
@@ -93,6 +93,24 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
...
                 "timeout_in_seconds": 3600
+            },
+
+            // Defines required OIDC related configuration
+            "oicd": {
+                // The ip address or hostname of the OIDC provider
+                "config_host": "oidc.example.com",
+
+                // The port the OIDC provider is listening on
+                "port": "8080",
```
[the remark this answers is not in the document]
MartinFlores751: True. I think initially I left it as a string because it's also used internally used as a string, so I didn't have to convert it.

Will change for consistency.

README.md
```diff
@@ -93,6 +93,24 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
...
+
+                // The path to the OIDC well-known endpoint
+                "uri": "/realms/example/.well-known/openid-configuration",
```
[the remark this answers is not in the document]
MartinFlores751: We could use `well_known_uri`, additionally, we could drop the `.well-known/openid-configuration` part since that is well known.

We could add something like the following:
> Retrieves configuration of OIDC server to contact the appropriate end-points.

README.md
```diff
@@ -93,6 +93,24 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
                 // The amount of time before a user's "Basic" authentication
                 // token expires.
                 "timeout_in_seconds": 3600
+            },
+
+            // Defines required OIDC related configuration
+            "oicd": {
+                // The ip address or hostname of the OIDC provider
```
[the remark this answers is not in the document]
korydraughn: I was mostly referring to switching the order of "ip address" and "hostname", as well as formatting text to look like the other locations.

We want to try to keep everything uniform as much as possible.

README.md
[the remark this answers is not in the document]
korydraughn: Yeah, I noticed why it was a string much later in the code review.

I think leaving it as a string is fine. The only concern with changing it to an integer is it could confuse admins.

README.md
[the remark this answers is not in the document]
korydraughn: Yeah, let's drop the parts that aren't required (especially for things that can be derived).

> We could add something like the following:
>
>> Retrieves configuration of OIDC server to contact the appropriate end-points.

Try to use phrasing that matches that of other properties. Overall, that wording seems fine. We'll work on it.

src/endpoints/authentication/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ target_include_directories(
   ${IRODS_EXTERNALS_FULLPATH_FMT}/include
   ${IRODS_EXTERNALS_FULLPATH_SPDLOG}/include
   CURL::libcurl
+  ${jwt_SOURCE_DIR}/include
```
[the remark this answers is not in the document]
MartinFlores751: Just tried `jwt` and `jwt::jwt`, doesn't seem to work... Renaming it to `jwt-cpp`, matching the name in the project's CMakeLists, allows us to do just that though.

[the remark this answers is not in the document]
korydraughn: What do you mean by **renaming**? What is **it**?

@SwooshyCueb Is there a simple way to determine what name should be used as the import target? (e.g. like `CURL::libcurl)`

Is it by convention? Are there rules for identifying this information?

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+		if (rc != 0) {
+			log::debug("Something happend....");
+		}
+
+		// Get service/port
+		//char *port{};
+		//rc = curl_url_get(endpoint, CURLUPART_PORT, &port, 0);
+		//if (rc != 0) {
+		//    log::debug("Something happend....");
+		//}
```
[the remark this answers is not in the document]
MartinFlores751: Will likely bring this section back to life, just wasn't sure how to handle the error case at the time, since we need a port.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
-        });
-    } // authentication
-} // namespace irods::http::endpoint
+	auto hit_token_endpoint(std::string encoded_body) -> nlohmann::json
+	{
+		const auto token_endpoint{
+			irods::http::globals::oidc_endpoint_configuration().at("token_endpoint").get<const std::string>()};
+
+		// Setup net
+		net::io_context io_ctx;
```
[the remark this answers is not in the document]
MartinFlores751: Would we end up putting it in the globals, or did you have something else in mind?

CMakeLists.txt
```diff
@@ -49,6 +49,17 @@ find_package(CURL REQUIRED)
 find_package(nlohmann_json "3.6.1" REQUIRED)
 find_package(OpenSSL REQUIRED COMPONENTS Crypto SSL)
 
+
+include(FetchContent)
+FetchContent_Declare(
+  jwt
+  GIT_REPOSITORY https://github.com/Thalhammer/jwt-cpp.git
```
SwooshyCueb, line 56: This appears to be a new dependency that is not provided by distribution package repositories. Any dependency that isn't or can't be provided by the distributions should be packaged as an externals package. Pulling things down from the internet at pre-build configure time is a no-go; at no point during the build process should internet access be required.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
-            res.set(field_type::content_type, "text/plain");
-            res.keep_alive(_req.keep_alive());
-            res.body() = std::move(bearer_token);
-            res.prepare_payload();
-
-            return _sess_ptr->send(std::move(res));
-        });
-    } // authentication
-} // namespace irods::http::endpoint
+	auto hit_token_endpoint(std::string encoded_body) -> nlohmann::json
```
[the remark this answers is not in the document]
MartinFlores751: What's the convention for leading underscores again?

SwooshyCueb (changes requested):

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+		const auto token_endpoint{
+			irods::http::globals::oidc_endpoint_configuration().at("token_endpoint").get<const std::string>()};
+
+		// Setup net
+		net::io_context io_ctx;
+		net::ip::tcp::resolver tcp_res{io_ctx};
+		beast::tcp_stream tcp_stream{io_ctx};
+
+		// Setup curl
+		CURLU* endpoint{curl_url()};
```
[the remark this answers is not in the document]
MartinFlores751: Think I followed what was given in the docs: https://curl.se/libcurl/c/curl_url.html

`NULL` is only returned when out of memory, not sure if that's something we can/should handle?

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+		return nlohmann::json::parse(res.body());
+	}
+
+	auto encode_string(std::string_view to_encode) -> std::string
+	{
+		// Init CURL
+		CURL* curl{curl_easy_init()};
+
+		// Ensure that CURL was successfully enabled
+		if (curl != nullptr) {
```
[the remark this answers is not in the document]
MartinFlores751: Turns out we don't need `curl_easy_init()` so we can remove it and the `nullptr` check.

Thinking of moving the `encode_string(...)` function to common, since it may be useful in other areas of the code base.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+
+		return std::transform_reduce(
+			std::next(std::begin(args)),
+			std::end(args),
+			encode_pair(*std::begin(args)),
+			[](auto a, auto b) { return a + "?" + b; },
+			encode_pair);
+	}
+
+	auto get_encoded_redirect_uri() -> std::string
```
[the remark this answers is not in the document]
MartinFlores751: Might remove this function. Think I created it to try out some ideas I had in mind at the time.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+						{"client_id",
+					     irods::http::globals::oidc_configuration().at("client_id").get_ref<const std::string&>()},
+						{"response_type", "code"},
+						{"scope", "openid"},
+						{"redirect_uri",
+					     irods::http::globals::oidc_configuration().at("redirect_uri").get_ref<const std::string&>()},
+						{"state", "placeholder"}};
+
+					const auto auth_endpoint{irods::http::globals::oidc_endpoint_configuration()
+					                             .at("authorization_endpoint")
```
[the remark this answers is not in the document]
MartinFlores751: `oidc_endpoint_configuration` is the OP (OpenID Provider) configuration gathered from the well-known endpoint, which we "hit" on server startup.

We know where to retrieve the `oidc_endpoint_configuration` information from the `uri` attribute(?) in the `oidc` section of the config.json file.

Maybe this means we might need a better name for this?

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+				                                       _req = std::move(_req),
+				                                       url = std::move(url)] {
+					// Two query params requiured by OAuth 2.0
+					// TODO: Double check with OIDC for bonus params
+					const auto code_iter{url.query.find("code")};
+					const auto state_iter{url.query.find("state")};
+
+					// Check to see if querys are valid
+					if (state_iter == std::end(url.query) || code_iter == std::end(url.query)) {
+						return _sess_ptr->send(fail(status_type::bad_request));
```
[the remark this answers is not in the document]
MartinFlores751: Currently, most OIDC error logs are being implemented in a separate branch. Will probably merge into this PR at some point.

src/endpoints/authentication/impl.cpp
MartinFlores751: Correct!

src/endpoints/authentication/impl.cpp
MartinFlores751: Think this was mentioned earlier, but we might be able to use the token timeouts that the OpenID Provider gives?

Haven't thought about the implications of going that route though. Would need to look more into it.

src/endpoints/authentication/impl.cpp
MartinFlores751: Likely artifact of the original OIDC grant implementation, before we used the 'iRODS' auth scheme.

Will remove.

src/endpoints/authentication/impl.cpp
MartinFlores751: Think you had changed this, but was lost in rebase. Will change to what you had.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
korydraughn: Use leading underscores for parameter names.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
korydraughn: I think you can extract the io_context from the socket.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
korydraughn: Let's check for null just to be on the safe side. It may help us in the future when things break.

src/endpoints/authentication/impl.cpp
korydraughn: Sure. We can handle that later though. Let's start with a simple config option in the OIDC config stanza.

That allows us to move faster.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+	{
+		auto encode_pair{
+			[](const BodyArguments::value_type& i) { return encode_string(i.first) + "=" + encode_string(i.second); }};
+
+		return std::transform_reduce(
+			std::next(std::begin(args)),
+			std::end(args),
+			encode_pair(*std::begin(args)),
+			[](auto a, auto b) { return a + "?" + b; },
+			encode_pair);
```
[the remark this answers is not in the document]
MartinFlores751: `encode_body(...)` uses `std::transform_reduce(...)` on `args` and first transforms `args` using `encode_pair(...)` to produce http encoded `key=value` parameters. Then, the transformed data is reduced using the lambda to build up the query string, e.g., `key=value?other=thing`.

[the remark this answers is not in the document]
korydraughn: Prefer using fmtlib to concat 3 or more std::strings.

`a + b + c` is a bad way to concat strings.

If you don't use fmtlib, you can use `+=` to concat strings. Don't forget to verify whether the args a and b are allowed to be modified.

Also, should `a` and `b` come in as `const auto&`?

README.md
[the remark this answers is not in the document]
MartinFlores751: Ended up changing it. Can always change back if needed.

src/endpoints/authentication/CMakeLists.txt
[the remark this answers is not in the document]
MartinFlores751: To clarify, changing the `FetchContent_Declare(...)`'s `jwt` to `jwt-cpp` allows use to use `jwt-cpp` instead of `${jwt_SOURCE_DIR}`. Can include in PR but I think the plan is to leave CMake alone until the external item is done.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
 
+#include <unordered_map>
+#include <algorithm>
+
+// clang-format off
+namespace beast = boost::beast; // from <boost/beast.hpp>
+namespace net   = boost::asio;  // from <boost/asio.hpp>
+// clang-format on
+
+using BodyArguments = std::unordered_map<std::string, std::string>;
```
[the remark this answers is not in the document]
MartinFlores751: Changed to `body_arguments`, but perhaps `query_arguments` might be a more fitting name, as it has to do with queries?

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
MartinFlores751: Added check with a horrible (temporary) debug message left for now. Need to think about what to log when we have no memory? Also need to improve the 'something happened' logs found below this line too...

src/endpoints/authentication/CMakeLists.txt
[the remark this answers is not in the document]
trel: > Can include in PR but I think the plan is to leave CMake alone until the external item is done.

That seems reasonable.   The intent is clear enough for now, and the external PR should be quick.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
MartinFlores751: The only two calls to this function pass an r-value to the arg, so it should be a move if I recall correctly.

We move the `encoded_body` later on, not sure if this would influence the decision on whether we capture by value in the case we receive an l-value?

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
MartinFlores751: Changed to `const auto &`, `std::begin` and `std::end` to const version.

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+
+					// Check to see if querys are valid
+					if (state_iter == std::end(url.query) || code_iter == std::end(url.query)) {
+						return _sess_ptr->send(fail(status_type::bad_request));
+					}
+
+					// Code here...
+					// Verify the state here!!!
+					log::debug("{}: Code is [{}]", fn, code_iter->second);
+					log::debug("{}: State is [{}]", fn, state_iter->second);
```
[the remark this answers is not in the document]
MartinFlores751: I think these logs are removed in another branch. These logs were more so to see I was holding things correctly.

Will likely remove.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
korydraughn: There may be a potential naming collision with that. Check common.hpp for the current set of type aliases. If there's no conflict, I don't see an issue.

src/endpoints/authentication/impl.cpp
```diff
@@ -246,10 +246,10 @@ namespace irods::http::handler
 					     irods::http::globals::oidc_configuration().at("redirect_uri").get_ref<const std::string&>()}};
 
 					// Encode the string, hit endpoint, get res
-					nlohmann::json res_item{hit_token_endpoint(encode_body(args))};
+					nlohmann::json oicd_response{hit_token_endpoint(encode_body(args))};
```
trel, line 249: oicd_response -> oidc_response

and then 3x below...

src/endpoints/authentication/CMakeLists.txt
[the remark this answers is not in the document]
SwooshyCueb: > @SwooshyCueb Is there a simple way to determine what name should be used as the import target? (e.g. like `CURL::libcurl)`
> 
> Is it by convention? Are there rules for identifying this information?

Import target names are usually defined by their find modules or cmake config files. The convention is `project::target` or `project::project`.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
korydraughn: I see. I think it's fine.

src/endpoints/authentication/impl.cpp
korydraughn: Capture "use OpenID timeouts" in an issue for now.

src/endpoints/authentication/impl.cpp
MartinFlores751: Hopefully got the spelling right this time...

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
MartinFlores751: Found a `query_arguments_type` which is the same type. Could use that, but not sure if it'll cause some confusion.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
MartinFlores751: Moved func to `encode(...)` in common. Forgot to prefix function arguments with `_` though...

src/endpoints/authentication/impl.cpp
korydraughn: If you've created the issue, please leave a comment here so we know where the issue came from.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
korydraughn: Ok. Feel free to mark resolved.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
korydraughn: Sounds confusing to me. If that alias is only used inside a function, consider moving it into the function's body where it is used.

src/endpoints/authentication/impl.cpp
```diff
@@ -18,122 +18,540 @@
...
+
+		return std::transform_reduce(
+			std::next(std::cbegin(_args)),
+			std::cend(_args),
+			encode_pair(*std::cbegin(_args)),
+			[](const auto& a, const auto& b) { return fmt::format("{}?{}", a, b); },
+			encode_pair);
+	}
+
+	auto is_error_response(const nlohmann::json& _response_to_check) -> bool
```
MartinFlores751, line 151: Not a big fan of the function name, could use something better.

src/endpoints/authentication/CMakeLists.txt
```diff
@@ -18,6 +19,7 @@ target_include_directories(
   ${IRODS_EXTERNALS_FULLPATH_BOOST}/include
   ${IRODS_EXTERNALS_FULLPATH_FMT}/include
   ${IRODS_EXTERNALS_FULLPATH_SPDLOG}/include
+  ${IRODS_EXTERNALS_FULLPATH_JWTCPP}/include
```
SwooshyCueb, line 22: Instead of using `IRODS_EXTERNALS_FULLPATH_JWTCPP`, add `jwt-cpp::jwt-cpp` to `target_link_libraries`. You may also be able to remove the `JWT_DISABLE_PICOJSON` (but probably not).

korydraughn: The `JWT_DISABLE_PICOJSON` macro was necessary for reaching a successful build.

Looking at the externals instructions for building the jwt package, seems we need to always define the macro.

Should it be part of `IRODS_COMPILE_DEFINITIONS_PUBLIC/PRIVATE` variables defined by irods/irods or is there a better way to automate that?

SwooshyCueb: Nah, it makes sense to put it here if it's needed.

korydraughn: Are there reasons to never define that macro?

SwooshyCueb: You wouldn't need to define it if it's passed down from `jwt-cpp::jwt-cpp`

korydraughn: Nice. @MartinFlores751 Please try using that instead.

SwooshyCueb: I don't remember if it is or not. Regardless, this line should still be removed from `target_include_directories` and `jwt-cpp::jwt-cpp` and `nlohmann_json::nlohmann_json` added to `target_link_libraries`. My original commit does this. I'm not sure why it didn't make it over to this PR.

korydraughn: The PR was rebased again. The changes you suggested should land again eventually.

src/endpoints/authentication/impl.cpp
```diff
@@ -180,8 +173,7 @@ namespace irods::http::handler
 		log::debug("{}: Authorization value (trimmed): [{}]", __func__, authorization);
 
 		constexpr auto max_creds_size = 128;
-		unsigned long size = max_creds_size;
-		//std::vector<std::uint8_t> creds(size);
+        std::uint64_t size{max_creds_size};
```
trel, line 176: indentation

src/endpoints/authentication/impl.cpp
```diff
@@ -388,16 +377,20 @@ namespace irods::http::handler
 					constexpr auto basic_auth_scheme_prefix_size = 6;
 					const auto [username, password]{decode_username_and_password(iter->value().substr(pos + basic_auth_scheme_prefix_size))};
 
+                    if (username.length() == 0 || password.length() == 0) {
```
trel, line 380: do we do length checks like this?   or do we have an `.empty()` method?

src/endpoints/authentication/impl.cpp
```diff
@@ -414,6 +407,9 @@ namespace irods::http::handler
 
 						int* correct{};
 
+                        // NOLINTNEXTLINE(cppcoreguidelines-owning-memory, cppcoreguidelines-no-malloc)
+                        irods::at_scope_exit free_memory{[&correct] { std::free(correct); }};
+
```
trel, line 412: indentation.  check your soft tabs or something... blaming the IDE...

src/endpoints/authentication/impl.cpp
SwooshyCueb: Looks like spaces were used instead of tabs

src/endpoints/authentication/impl.cpp
MartinFlores751: Yes, indentation is hideous. I'll need to either tweak some config or use clang-format(?).

@korydraughn not sure if clang-format is good to use now? Think we ignored it originally since we were playing with the config still?

src/endpoints/authentication/impl.cpp
MartinFlores751: Forgot about `empty()`...

src/main.cpp
```diff
@@ -339,6 +339,63 @@ auto init_irods_connection_pool(const json& _config) -> irods::connection_pool
...
 
+auto load_oidc_configuration(const json& _config) -> void
+{
+  nlohmann::json oi_config;
+        nlohmann::json endpoint_config;
+
+        // Scope as we're likely going to move this part out into a sep func...
+        try {
+			// Setup config
+            oi_config = _config.at(json::json_pointer{"/authentication/oidc"});
```
trel, line 350: indentation

src/main.cpp
```diff
@@ -339,6 +339,63 @@ auto init_irods_connection_pool(const json& _config) -> irods::connection_pool
...
+            // Consider reusing context further down main?
+            net::io_context io_ctx;
+            net::ip::tcp::resolver tcp_res{io_ctx};
+            beast::tcp_stream tcp_stream{io_ctx};
+
+            // Load config
+            const auto host{oi_config.at("config_host").get_ref<const std::string&>()};
+            const auto port{std::to_string(oi_config.at("port").get<int>())};
+            const auto temp_uri{oi_config.at("well_known_uri").get_ref<const std::string&>()};
+			const auto uri{fmt::format("{}/.well-known/openid-configuration", temp_uri)};
```
trel, line 362: yeah, need that sweet-sweet clang-format.

src/endpoints/authentication/impl.cpp
korydraughn: That's correct. We've been ignoring clang-format for now.

The plan was to run it one time across all source files so everything is updated in one commit. I am planning on going through the implementation and disabling clang-format for certain things since it can make code really ugly.

Tweak your editor's config to use whatever the project uses for indentation. You'll have to do that anyway if your editor is configured for spaces.

korydraughn: After some discussion, we've decided that @MartinFlores751 should run clang-format on his PR.

We can adjust formatting later if needed.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
MartinFlores751: We decided that we can leave this as `body_arguments` for now, and change later if need be.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
korydraughn: The session object will likely need to grow a new member function that exposes the `io_context`.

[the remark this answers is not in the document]
korydraughn: Let's ignore this for now. I'm concerned that we may have to touch more things to make this work.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
korydraughn: This is fine. Leave it as is.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
MartinFlores751: I'll create an issue for this, among other things 👍

src/endpoints/authentication/CMakeLists.txt
MartinFlores751: It should be put back in place now.

src/endpoints/authentication/impl.cpp
MartinFlores751: I'll be running clang-format, deferring formatting discussions to this thread(?): https://github.com/irods/irods_client_http_api/pull/75#discussion_r1334508464

src/main.cpp
MartinFlores751: Deferring formatting discussions to this thread(?): https://github.com/irods/irods_client_http_api/pull/75#discussion_r1334508464

src/main.cpp
MartinFlores751: Deferring formatting discussions to this thread(?): https://github.com/irods/irods_client_http_api/pull/75#discussion_r1334508464

src/endpoints/authentication/impl.cpp
```diff
@@ -20,120 +21,484 @@
...
+		// Send request
+		beast::http::write(tcp_stream, req);
+
+		// Read back req
+		beast::flat_buffer buffer;
+		beast::http::response<beast::http::string_body> res;
+		beast::http::read(tcp_stream, buffer, res);
+
+		log::debug("Got the following resp back: {}", res.body());
```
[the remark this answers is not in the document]
MartinFlores751: Moved to issue: #104

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
MartinFlores751: Moved over to #105

src/endpoints/authentication/impl.cpp
MartinFlores751: This should be covered by, if not highly related to, #77.

src/endpoints/authentication/impl.cpp
MartinFlores751: Similar to https://github.com/irods/irods_client_http_api/pull/75#discussion_r1294882373, it should be covered by #77.

src/main.cpp
korydraughn: Feel free to mark this and others like as resolved if linked to the same comment.

src/main.cpp
```diff
@@ -404,6 +405,62 @@ auto main(int _argc, char* _argv[]) -> int
...
+        // Confirm OIDC endpoint is valid (Assume all provide endpoint)
+        log::trace("Verifing OIDC endpoint configuration");
+
+        nlohmann::json oi_config;
+        nlohmann::json endpoint_config;
+
+        // Scope as we're likely going to move this part out into a sep func...
+        try {
+			// Setup config
+            oi_config = http_server_config.at(json::json_pointer{"/authentication/oidc"});
```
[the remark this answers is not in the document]
MartinFlores751: I think this would be tackled in #106 as part of the stricter config checking.

src/main.cpp
MartinFlores751: Deferring formatting discussions to this thread(?): https://github.com/irods/irods_client_http_api/pull/75#discussion_r1334508464

src/endpoints/authentication/impl.cpp
MartinFlores751: Formatted!

korydraughn:
Ohhh man, are we there!?

korydraughn:
Do all of the pytest/unittest tests pass?

SwooshyCueb:
Still seeing a lot of places where indentation is funky, or uses spaces instead of tabs, but the CMake changes look good

korydraughn:
The changes appear to be formatted correctly. The rest of the project uses spaces at the moment.

I'm going to run clang-format on the rest of the project once this PR is merged.

trel:
so close!

trel:
are we anticipating a great squashing?    so that it appears that martin got it all right on the first attempt?

if so, martin, you can select how many commits you want... pick what feels right.  could be 1.  could be 20.   make each one have purpose.

MartinFlores751:
> Do all of the pytest/unittest tests pass?

The tests did not pass initially, but I'm able to get the authentication tests to pass now.

Still, the tests in `test_irods_http_api.py` are failing, with the first failure reporting the following:
```
======================================================================
FAIL: test_common_operations (test_irods_http_api.test_data_objects_endpoint)
  File "/http_src/test/test_irods_http_api.py", line 496, in test_common_operations
    self.assertEqual(r.json()['irods_response']['error_code'], 0)
AssertionError: -78000 != 0

```

`ierror` tells us the error code is: `irods error: -78000 SYS_RESC_DOES_NOT_EXIST`


Still haven't looked too much into this issue yet, so I'm not sure what the cause is.

trel:
Check that server log.  It will tell you what it tried and failed to find.

MartinFlores751:
> Check that server log. It will tell you what it tried and failed to find.

Will do!

MartinFlores751:
All tests passed now!

trel:
congrats.

if no remaining open conversations on the PR... please rebase to taste when ready.

MartinFlores751:
Squashed.

trel:
do we need an issue for this work?

korydraughn:
We created an issue for this work a while back. @MartinFlores751 Please use that for your commits.

trel:
oh!  18.  indeed.

test/test_auth_endpoint.py
```diff
@@ -0,0 +1,115 @@
...
+# Pragmatic
+# Search for IRODS_HTTP_API_BASE_URL
+# and IRODS_CLIENT_VERSION
+
+@pytest.mark.parametrize("username, password, expected_result", [('bob', 'bob', requests.codes.ok),
+                                                                 ('', '', requests.codes.unauthorized),
+                                                                 ('not', 'valid', requests.codes.bad_request),
+                                                                 ('a'*200, 'b'*200, requests.codes.unauthorized)])
+def test_oidc_resource_owner_password_credentials_login(username, password, expected_result):
+    auth_url = 'http://127.0.0.1:9000/irods-http-api/0.1.0/authenticate'
```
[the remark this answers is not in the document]
MartinFlores751: Done!

trel (approved):
looks great.  this all passes and is ready?

korydraughn (approved):
If all tests pass, add the pounds.

MartinFlores751:
All tests passed, adding pounds.

PR #76, opened 2023-08-16
[irods/irods#7263] CMake consistency sweep

SwooshyCueb:


CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ set(IRODS_CLIENT_VERSION "0.1.0")
 set(IRODS_HTTP_API_VERSION "${IRODS_CLIENT_VERSION}")
 
 set(IRODS_PACKAGE_REVISION "1")
+set(IRODS_PACKAGE_EPOCH "1")
```
korydraughn, line 15: Don't think I've seen this before.

What does it mean?
Does it affect the package in any way?

korydraughn (approved):
Changes look fine.

Please add the pound.

CMakeLists.txt
SwooshyCueb: > Don't think I've seen this before.

You haven't. We haven't needed it before.

> What does it mean?
> Does it affect the package in any way?

It's part of the package version string. Every package without an epoch has an implied `0:` at the beginning of it. Setting the epoch just sets the number before the colon. Package epochs take precedent over every other part of the version string, and are typically only used when the versioning scheme changes. Here, the client version is independent from the irods version number, but the irods version number was still used in the call to `project()`.

HOWEVER, on a second look I don't think this is needed, as the package version is set explicitly later. I'll see if removing it breaks anything.

SwooshyCueb: Removed the epoch

SwooshyCueb:
\#'d

alanking (approved):

PR #82, opened 2023-09-05
Multiple Improvements

korydraughn:
Major updates:
- Ticket support for various operations
- Merging /metadata into various endpoints
- Atomic ACL modifications for /collections and /data-objects
- Connection pool refresh configuration options
- etc.

Something I noticed while writing the test for detecting resource changes was using seconds for the mtime does not provide enough precision for the connection pool to detect changes in all use-cases (e.g. very fast resource updates via a test).

Another thing that occurred to me while working on this was how to deal with the connection pool having a later mtime for a resource which no longer exists. I haven't determined whether we need to add logic for this case yet. If we do, that will be a change to the connection pool implementation in the irods/irods repo.

As of right now, this PR passes all tests consistently. I'm going to do another pass over it as there are a few more things I want to do before merging it.

I may wait until #75 is merged before merging this PR.

If interested, feel free to look through the PR and leave comments.


trel:
> how to deal with the connection pool having a later mtime for a resource which no longer exists.

oh, very interesting.     could... return the count of resources as well as the latest timestamp...  if timestamp goes up OR count goes down... then refresh.

CMakeLists.txt
```diff
@@ -60,6 +59,36 @@ endif()
...
+  # Make sure the correct llvm-symbolizer binary is available to Address Sanitizer. This binary
+  # allows debug symbols to be reported appropriately. There are two ways to do this:
+  #
+  #     export PATH=/opt/irods-externals/clang13.0.0-0/bin:$PATH
+  #
+  # - or -
+  #
+  #     export ASAN_SYMBOLIZER_PATH=/opt/irods-externals/clang13.0.0-0/bin/llvm-symbolizer
+  #
+  #add_compile_definitions(IRODS_ADDRESS_SANITIZER_DEFAULT_OPTIONS="log_path=/tmp/irods_client_http_api_asan_output:detect_odr_violation=1")
```
trel, line 76: how does this commented line align with the initial comment above?

README.md
```diff
@@ -172,7 +172,7 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
 
         // The credentials for the rodsadmin user that will act as a proxy
         // for all authenticated users.
-        "proxy_rodsadmin": {
+        "rodsadmin": {
```
trel, line 175: NFSRODS uses... 
```
    "proxy_admin_account": {
```

src/endpoints/CMakeLists.txt
```diff
@@ -1,11 +1,37 @@
...
+
+target_include_directories(
+  ${IRODS_HTTP_SHARED_API_OPERATIONS}
+  PRIVATE
+  ${IRODS_INCLUDE_DIRS}
+  ${IRODS_EXTERNALS_FULLPATH_BOOST}/include
+  ${IRODS_EXTERNALS_FULLPATH_FMT}/include
+  ${IRODS_EXTERNALS_FULLPATH_SPDLOG}/include
+  CURL::libcurl
+  #nlohmann_json::nlohmann_json
```
trel, line 24: is this coming soon?   or should be removed?

src/endpoints/authentication/CMakeLists.txt
```diff
@@ -25,13 +19,8 @@ target_include_directories(
   ${IRODS_EXTERNALS_FULLPATH_FMT}/include
   ${IRODS_EXTERNALS_FULLPATH_SPDLOG}/include
   CURL::libcurl
+  #nlohmann_json::nlohmann_json
```
trel, line 22: same

src/endpoints/data_objects/impl.cpp
```diff
@@ -300,6 +319,22 @@ namespace
...
+                            {"irods_response", {
+                                {"error_code", ec},
+                                {"error_message", "Error enabling ticket on connection."}
+                            }}
+                        }.dump();
+                        res.prepare_payload();
+                        return _sess_ptr->send(std::move(res));
+                    }
+                }
+
```
trel, line 337: this block of code is repeated... looks like 7-8x so far... any sane way to unify?

korydraughn:
Right. I'll capture it in an issue and get a fix in for the connection pool.

What are your thoughts on the mtime precision for resource updates? How do you feel about changing the modify_ts column to hold milliseconds? Or perhaps you feel adding a new column is more appropriate (not sure I like this even though it's safe).

I think all timestamp columns need to be changed to use milliseconds rather than seconds.

src/endpoints/data_objects/impl.cpp
```diff
@@ -465,12 +505,45 @@ namespace
...
+                            }.dump();
+                            res.prepare_payload();
+                            return _sess_ptr->send(std::move(res));
+                        }
+                    }
+
                     tp = std::make_unique<io::client::native_transport>(conn);
-                    out = std::make_unique<io::odstream>(*tp, lpath_iter->second);
+
+                    if (const auto iter = _args.find("resource"); iter != std::end(_args)) {
```
trel, line 540: do we need to consider hierarchies here?     probably already handled... but just leaving a note.

test.sh
```diff
@@ -464,6 +507,13 @@ test_tickets_endpoint()
 
 test_users_groups_endpoint()
 {
+    curl -H "authorization: Bearer $bearer_token" "${base_url}/users-groups" \
+        --data-urlencode 'op=modify_metadata' \
+        --data-urlencode 'name=kory' \
```
trel, line 512: let's remove kory from the test

test/test_irods_http_api.py
```diff
@@ -437,114 +575,175 @@ def test_common_operations(self):
...
+
+        # Read the contents of the data object and show it contains exactly what we expect.
+        r = requests.get(self.url_endpoint, headers=headers, params={
+            'op': 'read',
+            'lpath': data_object,
+            'count': 30
+        })
+        #print(r.content) # Debug
+        self.assertEqual(r.status_code, 200)
+        self.assertEqual(r.content.decode('utf-8'), 'A'* 10 + 'B' * 10 + 'C' * 10)
```
trel, line 675: space between 'A' and *

trel (commented):
saw a number of assertions commented out in the tests.

this is very impressive.

CMakeLists.txt
korydraughn: That is one of the things I'm planning on revisiting before merging this. So, the short answer is, I haven't spent the time thinking about it yet. I also haven't confirmed this works. Will do that soon though.

trel:
> I think all timestamp columns need to be changed to use milliseconds rather than seconds.

That's a really big change.   Many systems didn't support/report greater than 1-second precision for a long time.   So seconds was what was stored, and stored as epoch seconds because timezones are really hard.

We could look into it, but it would have to be a .0 type of thing, and maybe even 5.0.

I think we run with 1-second precision for now for this feature.

README.md
korydraughn: Right. I made this change before I realized that.

Do we want to use the NFSRODS name?

CMakeLists.txt
trel: Gotcha.   Resolve whenever you feel ready.

README.md
trel: yes please - it still looks good and is clean/clear.

src/endpoints/CMakeLists.txt
korydraughn: I think that is what we want, but it doesn't work for some reason. I'm going to look into it to try and learn why before I reach out to @SwooshyCueb.

src/endpoints/data_objects/impl.cpp
korydraughn: Yes, there is. I'm going to take care of that soon.

src/endpoints/data_objects/impl.cpp
korydraughn: So far, none of the operations support hierarchies. That's not to say they won't, no thought has been spent on supporting them yet.

Willing to discuss whether they should be supported though.

src/endpoints/data_objects/impl.cpp
trel: ah, very nice.

test.sh
korydraughn: This file may go away entirely since the official test suite is written in python.

src/endpoints/data_objects/impl.cpp
trel: no interest in supporting them.   just want to make sure this code works as desired when handed one.   aka, does it error out correctly/as expected?

korydraughn:
Switching to millisecond precision is a no go due to the values becoming too large to represent over time. However, we can introduce a new column that holds the fractional part of the mtime.

The new column could be in milliseconds or even microseconds and initialized to 0 making it completely safe.

src/endpoints/CMakeLists.txt
SwooshyCueb: We don't want to put targets in `target_include_directories` at all. They go in `target_link_libraries`.

SwooshyCueb (changes requested):

src/endpoints/CMakeLists.txt
korydraughn: The commented out code certainly didn't work. Things only compiled when using `${IRODS_EXTERNALS_FULLPATH_JSON}/include`.

I'm guessing we aren't doing what CURL is doing then? I feel like I've seen `nlohmann_json::nlohmann_json` in an include directive before.

SwooshyCueb: `CURL::libcurl` should be in `target_link_libraries`, too. Passing a target to `target_link_libraries` takes care of includes as well as any library linkages. You shouldn't need to *ever* reference `IRODS_EXTERNALS_FULLPATH_JSON` unless something has gone horribly wrong.

korydraughn: Oh. I had no idea `target_link_libraries` handled include paths. I'll remove that and see what happens.

src/endpoints/data_objects/impl.cpp
korydraughn: I'd expect either an error or undefined behavior in the context of iRODS.

Also, dstream doesn't support hierarchies.

trel: fair enough - resolving.

src/endpoints/data_objects/impl.cpp
korydraughn: The more I look at these, the more I feel it is fine.

The only lines that seem worthy of unifying are the ones dealing with constructing an error response.

If we try and encapsulate the inner most if-statement, we either have to throw an exception which isn't ideal (since we can just return) or we have to return more info from the new function.

I'm going to consolidate the body of the inner if-statement.

korydraughn: Created issue #85 for this.

Leaving as is for now.

README.md
korydraughn: Done.

src/endpoints/CMakeLists.txt
korydraughn: Done.

src/endpoints/authentication/CMakeLists.txt
korydraughn: Done.

test.sh
korydraughn: Created issue #86 to address this.

test/test_irods_http_api.py

src/endpoints/query/impl.cpp
```diff
@@ -162,6 +162,21 @@ namespace
...
+                        res.body() = json{
+                            {"irods_response", {
+                                {"error_code", ec},
+                                {"error_message", "Error enabling ticket on connection."}
+                            }}
+                        }.dump();
+                        res.prepare_payload();
+                        return _sess_ptr->send(std::move(res));
+                    }
+                }
```
korydraughn, line 178: Should the query endpoint accept tickets?

`iquest` doesn't expose a way to enable tickets, but it could. Perhaps there's a reason behind it not exposing a ticket option.

src/endpoints/shared_api_operations.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+#include "globals.hpp"
+#include "log.hpp"
+#include "session.hpp"
+#include "version.hpp"
+
+#include <irods/atomic_apply_acl_operations.h>
+#include <irods/atomic_apply_metadata_operations.h>
+#include <irods/filesystem.hpp>
+
```
korydraughn, line 11: Remove empty line.

src/endpoints/query/impl.cpp
korydraughn: After some discussion, we've concluded the /query endpoint should not accept tickets.

Only the following operations should accept tickets.
- Stat'ing a data object
- Stat'ing a collection
- Writing to a data object
- Reading a data object

korydraughn: Done

src/endpoints/shared_api_operations.cpp
korydraughn: Done

CMakeLists.txt
korydraughn: Done

korydraughn:
All comments have been addressed.

All tests pass with ASan enabled!

Squashing.

korydraughn:
This PR is in a good spot for merging.

Taking out of draft mode.

trel:
How much does it conflict with #75?     Let's merge ASAP if you're good with it.   Fewer dangling pieces out there.

korydraughn:
There shouldn't be too much conflict since #75 is dealing with auth stuff. This PR touches the auth endpoint, but only for trivial changes.

trel:
Agreed - if it rebases cleanly in a test branch, we'll know for sure.

korydraughn:
This PR touches the auth endpoint in the following ways:
- Fixes a memory leak
- Renames the `rodsadmin` config property to `proxy_admin_account`
- Tweaks the CMakeLists.txt file

trel:
Okay, easy enough to see then...

Yes, please merge this one when #'d and ready.

korydraughn:
Added #'s

PR #103, opened 2023-09-18
4.2 compatibility and other tweaks

korydraughn:


src/common.cpp
```diff
@@ -240,20 +242,23 @@ namespace irods
         using irods::experimental::filesystem::perms;
 
         switch (_p) {
+            // clang-format off
             case perms::null:            return "null";
             case perms::read_metadata:   return "read_metadata";
-            case perms::read_object:     return "read_object";
-            case perms::read:            return "read_object";
+            case perms::read_object:
+            case perms::read:            return "read";
```
trel, line 249: still feeling like we can wrap these two (and the two for `write`) in a `enable_4_2_compatibility` conditional...

```
if enable_4_2_compatibility
    read
else
    read_object
```

src/common.cpp
```diff
@@ -263,13 +268,13 @@ namespace irods
         // clang-format off
         if (_s == "null")            { return perms::null; }
         if (_s == "read_metadata")   { return perms::read_metadata; }
-        if (_s == "read_object")     { return perms::read_object; }
-        if (_s == "red")             { return perms::read; }
```
trel, line 267: whoa, just a straight typo?!

src/endpoints/authentication/impl.cpp
```diff
@@ -82,33 +83,47 @@ namespace irods::http::handler
...
+                static const auto& config = irods::http::globals::configuration();
+                static const auto& rodsadmin_username = config.at(json_pointer{"/irods_client/proxy_admin_account/username"}).get_ref<const std::string&>();
+                static const auto& rodsadmin_password = config.at(json_pointer{"/irods_client/proxy_admin_account/password"}).get_ref<const std::string&>();
+                static const auto& zone = config.at(json_pointer{"/irods_client/zone"}).get_ref<const std::string&>();
 
-                namespace adm = irods::experimental::administration;
-                const adm::user_password_property prop{password, rodsadmin_password};
-                const auto obfuscated_password = irods::experimental::administration::obfuscate_password(prop);
-                obfuscated_password.copy(input.password, sizeof(CheckAuthCredentialsInput::password));
+                if (config.at(json_pointer{"/irods_client/enable_4_2_compatibility"}).get<bool>()) {
```
trel, line 93: add a couple comments on what we have to  / will do differently in this block.

test/config.py
```diff
@@ -14,6 +14,7 @@
     },
 
     'irods_zone': 'tempZone',
+    'irods_hostname_for_resources': '97f627fe73a6',
```
trel, line 17: seems... particular.    do we want a hostname here?

korydraughn: This was added due to the tests needing to create resources in the docker container running iRODS.

This may not be necessary, but I'm pretty sure it is.

src/common.cpp
trel: Agreed on the apps *using* the HTTP API... but I think we optimize for 4.3+.   This 4.2 layer is a workaround/compromise for a couple years until everyone is on 4.3+.

I'd rather have a box around the 'old' names and use the new names wherever possible.   It's only not possible here everywhere because 4.2 doesn't understand the new ones, so we make a workaround until they upgrade.  We can write that somewhere in the README for this compatibility layer so a developer/admin can make an appropriate assessment about where they want to spend their effort/time.

test/config.py
trel: okay, just caught my eye as surprising.

src/common.cpp
korydraughn: With a little bit of work, we can make the HTTP API accept both forms of input and present only the 4.3 names when sending a response.

test/config.py
korydraughn: I'm not surprised. I was not happy when I added that property, but I don't know if there's a better way to do it yet.

trel: can we just inject it dynamically?

src/common.cpp
trel: that could work.   any app that uses this will, by definition, be new... so, that seems okay to always send the future-proof strings back.

test/config.py
trel: right, because you're removed from the point of activity.    okay... will keep thinking.   carry on.

src/common.cpp
korydraughn: The application now accepts 4.3 and 4.2 names for permissions, but only returns the 4.3 names.

src/endpoints/authentication/impl.cpp
korydraughn: done

test/config.py
korydraughn: done

korydraughn:
I've also confirmed that all tests pass when run against a 4.2.11 server.

src/common.hpp
```diff
@@ -70,6 +72,49 @@ namespace irods::http
         const authenticated_client_info* client_info{};
     }; // struct client_identity_resolution_result
 
+    class irods_connection // NOLINT(cppcoreguidelines-special-member-functions)
```
korydraughn, line 75: Come up with a better name.
- `connection_facade`
- `connection_adapter`

This class needs to be part of the iRODS server. That can happen in a later release though. _(Capture this in an issue)_

korydraughn: Renamed to `connection_facade`.

korydraughn:
Ignore clang-format workflow.

Formatting will be addressed in a separate PR after this is merged.

korydraughn:
This PR is ready for merging.

README.md
```diff
@@ -188,6 +188,20 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
             "verify_server": "cert"
         },
 
+        // Controls how the HTTP API communicates with the iRODS server.
+        //
+        // When set to true, the following applies:
+        // - Only APIs supported by iRODS 4.2.11 will be used.
```
trel, line 194: no mention of 4.2.12... is that intentional?

and if someone is running 4.2.10 or prior?

src/common.cpp
```diff
@@ -254,20 +256,23 @@ namespace irods
...
-            case perms::modify_object:   return "modify_object";
+            case perms::modify_object:
             case perms::write:           return "modify_object";
             case perms::delete_object:   return "delete_object";
             case perms::own:             return "own";
-            default:                     return "?"; // TODO std::unreachable() or __builtin_unreachable()
+            // clang-format on
         }
+
+        THROW(SYS_INVALID_INPUT_PARAM, fmt::format("Cannot convert permission enumeration to string."));
```
trel, line 275: any way to write/show what it tried to do?

src/common.hpp
```diff
@@ -70,6 +72,49 @@ namespace irods::http
...
+            if (auto* p = std::get_if<irods::connection_pool::connection_proxy>(&conn_); p) {
+                return *p;
+            }
+
+            return *std::get_if<irods::experimental::client_connection>(&conn_);
+        } // operator RcComm&
+
+      private:
+        std::variant<std::monostate, irods::experimental::client_connection, irods::connection_pool::connection_proxy> conn_;
+    }; // class irods_connection
```
trel, line 116: connection_facade

src/endpoints/resources/impl.cpp
```diff
@@ -498,7 +499,45 @@ namespace
...
+                                                 "RESC_CONTEXT, RESC_COMMENT, RESC_INFO, "
+                                                 "RESC_FREE_SPACE, RESC_FREE_SPACE_TIME, "
+                                                 "RESC_PARENT, RESC_CREATE_TIME, RESC_MODIFY_TIME "
+                                                 "where RESC_NAME = '{}'",
+                                                 name_iter->second);
+
+                    for (auto&& row : irods::experimental::query_builder{}.build<RcComm>(conn, gql)) {
+                        exists = true;
+
+                        info = { // TODO Can't use += yet
```
trel, line 516: we have an issue for this?

README.md
korydraughn: Oh, missed that. It's supposed to specify the minimum version tested.

4.2.11 and later.

src/common.cpp
korydraughn: Yes. Will adjust.

src/endpoints/resources/impl.cpp
korydraughn: We don't. That TODO is from the irods/irods code.

Will remove since it's not required.

src/common.cpp
korydraughn: Actually, there's no way to do that since we're given an enumeration.

We should never reach the `THROW` unless we do something suspicious like cast an integer to an enumeration.

README.md
korydraughn: Done.

src/common.hpp
korydraughn: Done.

src/endpoints/resources/impl.cpp
korydraughn: Removed.

trel (approved):
Looks good.  Let’s pound them and get them in.

korydraughn:
Added #'s

PR #116, opened 2023-10-25
`process_stash` fix

MartinFlores751:
This should fix the potential `process_stash` issues stated in irods/irods#7297.
This solution just returns a copy of the object desired, rather than providing a pointer.

All existing tests pass with the changes.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -2,12 +2,12 @@
 
 #include "globals.hpp"
 #include "log.hpp"
+#include "process_stash.hpp"
 #include "version.hpp"
 
 #include <irods/client_connection.hpp>
 #include <irods/irods_at_scope_exit.hpp>
 #include <irods/irods_exception.hpp>
-#include <irods/process_stash.hpp>
 #include <irods/rcConnect.h>
 #include <irods/rcMisc.h> // For addKeyVal().
 #include <irods/rodsErrorTable.h>
@@ -232,20 +232,25 @@ namespace irods::http
         log::debug("{}: Bearer token: [{}]", __func__, bearer_token);
 
         // Verify the bearer token is known to the server. If not, return an error.
-        const auto* object = irods::process_stash::find(bearer_token);
-        if (!object) {
-            log::error("{}: Could not find bearer token matching [{}].", __func__, bearer_token);
-            return {.response = fail(status_type::unauthorized)};
+        auto mapped_value{irods::http::process_stash::find(bearer_token)};
+		if (!mapped_value.has_value()) {
+		  log::error("{}: Could not find bearer token matching [{}].", __func__, bearer_token);
+		  return {.response = fail(status_type::unauthorized)};
         }
 
-        const auto* p = boost::any_cast<authenticated_client_info>(&*object);
-        if (std::chrono::steady_clock::now() >= p->expires_at) {
-            log::error("{}: Session for bearer token [{}] has expired.", __func__, bearer_token);
-            return {.response = fail(status_type::unauthorized)};
+		auto* client_info{boost::any_cast<authenticated_client_info>(&*mapped_value)};
+		if (client_info == nullptr) {
+		  log::error("{}: Could not find bearer token matching [{}].", __func__, bearer_token);
+		  return {.response = fail(status_type::unauthorized)};
         }
 
-        log::trace("{}: Client is authenticated.", __func__);
-        return {.client_info = p};
+		if (std::chrono::steady_clock::now() >= client_info->expires_at) {
+		  log::error("{}: Session for bearer token [{}] has expired.", __func__, bearer_token);
+		  return {.response = fail(status_type::unauthorized)};
+		}
+
+		log::trace("{}: Client is authenticated.", __func__);
+		return {.client_info = std::move(*client_info)};
     } // resolve_client_identity
 } // namespace irods::http
 
--- a/src/process_stash.cpp
+++ b/src/process_stash.cpp
@@ -1,0 +1,84 @@
+#include "process_stash.hpp"
+
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_generators.hpp>
+#include <boost/uuid/uuid_io.hpp>
+
+#include <iterator>
+#include <optional>
+#include <shared_mutex>
+#include <unordered_map>
+
+namespace
+{
+    // A mapping containing handles to heterogenous objects.
+    std::unordered_map<std::string, boost::any> g_stash; // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
+
+    // A mutex which protects the map from data corruption.
+    std::shared_mutex g_mtx; // NOLINT(cppcoreguidelines-avoid-non-const-global-variables, cert-err58-cpp)
+
+    auto generate_unique_key() -> std::string
+    {
+        std::string uuid;
+        constexpr auto uuid_length{36};
+        uuid.reserve(uuid_length);
+        uuid = to_string(boost::uuids::random_generator{}());
+
+        while (g_stash.find(uuid) != std::end(g_stash)) {
+            uuid = to_string(boost::uuids::random_generator{}());
+        }
+
+        return uuid;
+    } // generate_unique_key
+} // anonymous namespace
+
+namespace irods::http::process_stash
+{
+    auto insert(boost::any _value) -> std::string
+    {
+        std::lock_guard lock{g_mtx};
+        return g_stash.insert_or_assign(generate_unique_key(), std::move(_value)).first->first;
+    } // insert
+
+    auto find(const std::string& _key) -> std::optional<boost::any>
+    {
+        {
+            std::shared_lock lock{g_mtx};
+            if (auto iter = g_stash.find(_key); iter != std::end(g_stash)) {
+                return iter->second;
+            }
+        }
+
+        return std::nullopt;
+    } // find
+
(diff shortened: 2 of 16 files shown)


src/common.cpp
```diff
@@ -232,20 +231,20 @@ namespace irods::http
...
+        const auto object{irods::http::process_stash::find(bearer_token)};
+		try {
+		  const auto p{boost::any_cast<const authenticated_client_info&>(object)};
+		  if (std::chrono::steady_clock::now() >= p.expires_at) {
+            log::error("{}: Session for bearer token [{}] has expired.", __func__, bearer_token);
             return {.response = fail(status_type::unauthorized)};
-        }
+		  }
+          log::trace("{}: Client is authenticated.", __func__);
+		  return {.client_info = p};
```
MartinFlores751, line 242: I think if we were to not `const` the variable `p`, we could do a move on the final line?

src/process_stash.cpp
```diff
@@ -0,0 +1,82 @@
...
+    auto find(const std::string& _key) -> boost::any
+    {
+        {
+            std::shared_lock lock{g_mtx};
+            if (auto iter = g_stash.find(_key); iter != std::end(g_stash)) {
+                return iter->second;
+            }
+        }
+
+        return nullptr;
```
MartinFlores751, line 50: Not sure how we feel about using a `boost::any(nullptr)` to signify a failure to find a key?

MartinFlores751 (commented):
Just a few things I wanted to note down.

alanking:
Are we anticipating further changes? Just wondering why this is in draft.

src/common.cpp
```diff
@@ -232,20 +231,20 @@ namespace irods::http
         log::debug("{}: Bearer token: [{}]", __func__, bearer_token);
 
         // Verify the bearer token is known to the server. If not, return an error.
-        const auto* object = irods::process_stash::find(bearer_token);
-        if (!object) {
-            log::error("{}: Could not find bearer token matching [{}].", __func__, bearer_token);
+        const auto object{irods::http::process_stash::find(bearer_token)};
```
alanking, line 234: Since we are touching these lines anyway, let's make the tabs and spaces consistent in this area so it's a little easier to read

alanking, line 234: Can we name this something to make it a little more indicative of what it is?

alanking (commented):
Had some naive thoughts :)

Please make sure to leave a detailed description in the commit message and/or in the header files of the differences between the "official" process stash and this one.

Is the plan to port these changes to the official one and drop the `irods::http` implementation?

MartinFlores751:
> Are we anticipating further changes? Just wondering why this is in draft.

Shouldn't be anything fundamental? Will change over to non-draft.

src/common.cpp
MartinFlores751: Can do, but I think the plan at some point was that @korydraughn was going to do a formatting sweep on the entire repo?

I remember the formatting was a bit odd on the OIDC code too...

trel: Yes, as long as it's about right, that's fine - there will be a full sweep prior to release.

alanking: Yep, just took a little more brain power than usual to keep the scopes straight in my head ;)

src/common.cpp
korydraughn: Go for it.

src/common.cpp
```diff
@@ -1,13 +1,12 @@
 #include "common.hpp"
-
```
korydraughn, line 2: Please restore this empty line.

That helps the reader understand that this cpp file is the implementation of the first header file.

src/process_stash.cpp
korydraughn, line 50: Based on the documentation, we have to check for `nullptr` now.

Is the use of `find` throughout the codebase safe? Do we need more checks?

korydraughn (commented):
Very nice.

src/common.cpp
MartinFlores751: Should be a bit better now?

src/common.cpp
```diff
@@ -232,20 +231,20 @@ namespace irods::http
         log::debug("{}: Bearer token: [{}]", __func__, bearer_token);
 
         // Verify the bearer token is known to the server. If not, return an error.
-        const auto* object = irods::process_stash::find(bearer_token);
-        if (!object) {
-            log::error("{}: Could not find bearer token matching [{}].", __func__, bearer_token);
+        const auto object{irods::http::process_stash::find(bearer_token)};
+		try {
+		  const auto p{boost::any_cast<const authenticated_client_info&>(object)};
```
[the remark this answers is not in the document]
MartinFlores751: Name should be clearer now.

src/process_stash.cpp
MartinFlores751: Don't think you should be able to access `nullptr` unless you do the following:

`boost::any_cast<std::nullptr_t>(null_bad);`

Could be wrong, might need more testing.
Either way, wanted to bring up since `nullptr` being there is kinda odd, even if it works.

korydraughn: Yes, I agree. Seeing `nullptr` there definitely raised an eyebrow or two.

korydraughn: Consider `std::optional<boost::any>`.

korydraughn:
Once this PR lands, I'm going to format the entire project.

src/process_stash.cpp
MartinFlores751: Used `std::optional` approach.

src/process_stash.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Searches the process stash for the value associated with a specific key.
+    ///
+    /// This function is thread-safe.
+    ///
+    /// \param[in] _key The string which maps to the value of interest.
+    ///
+    /// \returns A pointer to the value of interest.
+    /// \retval A boost::any object if the key exists.
+    /// \retval A boost::any object containing a nullptr otherwise.
```
alanking, line 43: I believe this is now `std::nullopt`

src/process_stash.hpp
```diff
@@ -0,0 +1,89 @@
...
+    ///
+    /// \since 4.2.12
```
alanking, line 45: Just making a note here vis-a-vis porting this back to irods/irods...

Do we need to follow our deprecation model for this (or any other function) since we are changing the interface and this is client-facing? I realize it's a relatively new, C++ client library, so maybe we won't need to, but just wanted to pose the question.

Should we follow the deprecation model, I guess we would need a differently named function at first (since you can't overload based on return type) and then do a monkey-head later to get back to the desired function name.

alanking (commented):
Changes seem good. Need a Doxygen tweak and had a question about deprecation that we don't need to answer right now

src/process_stash.hpp
korydraughn: I don't think we need to follow a deprecation model for this. The interfaces are small enough to eyeball and spot differences.

It's also possible that this be merged into irods/irods as an alternative to the existing process stash.

Let's let it ride for a while and see what it becomes.

src/process_stash.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+namespace irods::http::process_stash
+{
+    auto insert(boost::any _value) -> std::string
+    {
+        std::lock_guard lock{g_mtx};
+        return g_stash.insert_or_assign(generate_unique_key(), std::move(_value)).first->first;
+    } // insert
+
+  auto find(const std::string& _key) -> std::optional<boost::any>
```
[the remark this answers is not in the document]
MartinFlores751: Done!

src/process_stash.hpp
MartinFlores751: Done!

src/process_stash.hpp
korydraughn: Resolving.

korydraughn:
Do the tests pass?

Are there any additional things left to do for this PR?

MartinFlores751:
That should be all, I'll rerun the tests just to be sure though.

korydraughn:
If the tests pass, squash to taste and leave off the pounds.

MartinFlores751:
Should this refer to the irods/irods issue, or should we create one for the HTTP API?

korydraughn:
Let's create an issue for this repo.

korydraughn:
Sorry, let's use the issue in irods/irods.

MartinFlores751:
All tests past 3 times in row & tokens say they got evicted properly, going to squash.

korydraughn (approved):
Please leave a comment after force-pushing so we're notified. GitHub doesn't notify us about force-pushes.

Pound it.

MartinFlores751:
Pounded!

src/process_stash.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <boost/uuid/uuid_io.hpp>
+
+#include <iterator>
+#include <optional>
+#include <shared_mutex>
+#include <unordered_map>
+
+namespace
+{
+    // A mapping containing handles to heterogenous objects.
```
trel, line 14: heterogenous -> heterogeneous

trel: no worries - can tweak it now or later.

alanking: Later!

trel: oh, it was already in the irods/irods version... ignore me!

Issue #113, opened 2023-10-19
Add operation that allows rodsadmins to modify properties of replicas


korydraughn:
This is essentially about exposing the functionality of `iadmin modrepl`.

PR #118, opened 2023-10-27
[#113] Add support for modifying replica properties

korydraughn:
Passes tests.

### DIFF
--- a/API.md
+++ b/API.md
@@ -38,7 +38,7 @@ curl http://localhost:<port>/irods-http-api/<version>/collections \
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -70,7 +70,7 @@ If `recurse` is set to 1, the contents of the collection will be removed. If `no
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -102,7 +102,7 @@ If `ticket` is passed a valid ticket string, it will be enabled before carrying 
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -145,7 +145,7 @@ curl http://localhost:<port>/irods-http-api/<version>/collections \
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -181,7 +181,7 @@ curl http://localhost:<port>/irods-http-api/<version>/collections \
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -224,7 +224,7 @@ The JSON object passed to the `operations` parameter must have the following str
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -286,7 +286,7 @@ The JSON object passed to the `operations` parameter must have the following str
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -334,7 +334,7 @@ curl http://localhost:<port>/irods-http-api/<version>/collections \
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -369,7 +369,7 @@ curl http://localhost:<port>/irods-http-api/<version>/data-objects \
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -401,7 +401,7 @@ curl http://localhost:<port>/irods-http-api/<version>/data-objects \
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
@@ -436,7 +436,7 @@ curl http://localhost:<port>/irods-http-api/<version>/data-objects \
 
 #### Response
 
-If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
 
 ```js
 {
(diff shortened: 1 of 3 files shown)


korydraughn:
TODO: Allow multiple properties of a replica to be updated at the same time. This is already supported by iRODS. This operation just needs to adjust its interface.

alanking (commented):
Seems good overall. Might want to consider adding tests for any/all failure modes you can think of as it pertains to the endpoint (i.e. don't need to test `rcModDataObjMeta`).

korydraughn:
Agreed. Will cover failure cases later.

Created #119 for that effort.

test/test_irods_http_api.py
```diff
@@ -848,6 +848,65 @@ def test_modifying_permissions_atomically(self):
...
+        result = r.json()
+        self.assertEqual(result['irods_response']['error_code'], 0)
+        self.assertEqual(result['rows'][0][0], '0')
+
+        # Remove the data object.
+        r = requests.post(self.url_endpoint, headers=headers, data={
+            'op': 'remove',
+            'lpath': data_object,
+            'no-trash': 1
+        })
```
alanking, line 905: Any reason this isn't in a `finally` block?

src/endpoints/data_objects/impl.cpp
```diff
@@ -1845,4 +1851,132 @@ namespace
...
+
+                if (const auto name_iter = _args.find("property-name"); name_iter != std::end(_args)) {
+                    static constexpr auto properties = std::to_array<std::pair<const char*, const char*>>({
+                        {"COLL_ID", COLL_ID_KW},
+                        {"DATA_CREATE_TIME", DATA_CREATE_KW},
+                        {"DATA_CHECKSUM", CHKSUM_KW},
+                        {"DATA_EXPIRY", DATA_EXPIRY_KW},
+                        {"DATA_ID", DATA_ID_KW},
+                        {"DATA_REPL_STATUS", REPL_STATUS_KW},
+                        {"DATA_MAP_ID", DATA_MAP_ID_KW},
```
alanking, line 1906: Just noting for the future that DATA_ID and DATA_MAP_ID cannot be modified through the API (enforced by the database plugin).

If we want this to mirror `iadmin modrepl`, we need to disallow modifying the following columns:
 - COLL_ID
 - DATA_ID
 - DATA_MAP_ID
 - DATA_NAME
 - DATA_RESC_HIER
 - DATA_RESC_NAME

Arguably, the API should enforce this, but... I don't think that it does. Also, that would require some admin-only shenanigans as these columns do need to be modified for legitimate cases. Modifying these columns outside of those legitimate cases can really put you in a situation.

I think at the very least that we need to note somewhere that modifying these columns is dangerous and may not be recoverable.

trel: Just noting that is not enough I feel.   We need to enforce it - real bumpers.

Feels like those bumpers just being here for now (same as in `iadmin modrepl`) is sufficient to me...

test/test_irods_http_api.py
korydraughn: It's because the testing is still a WIP. Making everything clean is on the TODO list, but that is going to happen at a later time.

src/endpoints/data_objects/impl.cpp
korydraughn: Will do.

test/test_irods_http_api.py
```diff
@@ -848,6 +848,65 @@ def test_modifying_permissions_atomically(self):
...
+    def test_modifying_replica_properties(self):
+        headers = {'Authorization': 'Bearer ' + self.rodsadmin_bearer_token}
+
+        # Create a data object.
+        data_object = os.path.join('/', self.zone_name, 'home', self.rodsadmin_username, 'modrepl.txt')
+        r = requests.post(self.url_endpoint, headers=headers, data={
+            'op': 'touch',
+            'lpath': data_object
+        })
+        #print(r.content) # Debug
```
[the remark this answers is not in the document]
korydraughn: That's a good point. Will create an issue to address that in a later PR.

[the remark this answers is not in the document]
korydraughn: Created #123.

src/endpoints/data_objects/impl.cpp
```diff
@@ -1845,4 +1851,132 @@ namespace
...
     } // op_modify_metadata
+
+    IRODS_HTTP_API_ENDPOINT_OPERATION_SIGNATURE(op_modify_replica)
+    {
+        auto result = irods::http::resolve_client_identity(_req);
+        if (result.response) {
+            return _sess_ptr->send(std::move(*result.response));
+        }
+
+        const auto* client_info = result.client_info;
```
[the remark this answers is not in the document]
korydraughn: Will do.

src/endpoints/data_objects/impl.cpp
korydraughn: I'm considering the removal of DATA_OWNER_NAME and DATA_OWNER_ZONE. Since ichmod no longer treats them special, it feels right to not allow people to modify them. I don't think there's a valid use-case for allowing those to be modified now.

trel: I think i agree.   do we allow those for modrepl?

korydraughn: Yes.

trel: oh, hmmm.

korydraughn: Ok. I'm going to remove them. We can always add them back in if needed.

src/endpoints/data_objects/impl.cpp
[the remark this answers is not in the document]
korydraughn: Done.

src/endpoints/data_objects/impl.cpp
korydraughn: Done.

API.md
```diff
@@ -971,6 +971,54 @@ If there was an error, expect either an HTTP status code in the 4XX or 5XX range
...
+    --data-urlencode 'new-data-mode=<string>' \
+    --data-urlencode 'new-data-modify-time=<string>' \
+    --data-urlencode 'new-data-path=<string>' \
+    --data-urlencode 'new-data-replica-number=<integer>' \
+    --data-urlencode 'new-data-replica-status=<integer>' \
+    --data-urlencode 'new-data-resource-id=<integer>' \
+    --data-urlencode 'new-data-size=<integer>' \
+    --data-urlencode 'new-data-status=<string>' \
+    --data-urlencode 'new-data-type-name=<string>' \
+    --data-urlencode 'new-data-version=<string>'
```
korydraughn, line 1000: Any thoughts on this?

API.md
```diff
@@ -971,6 +971,54 @@ If there was an error, expect either an HTTP status code in the 4XX or 5XX range
...
+```
+
+`resource-hierarchy` and `replica-number` are mutually exclusive parameters.
+
+All parameters having a prefix of `new-` represent modifiable properties of the target replica.
+
+#### Response
+
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. It's structure is shown below.
```
korydraughn, line 1009: Will fix the documentation for the other APIs later.

`It's` -> `Its`.

korydraughn:
All tests pass.

API.md

src/endpoints/data_objects/impl.cpp
```diff
@@ -1796,4 +1803,120 @@ namespace
...
+					{"new-data-replica-status", REPL_STATUS_KW},
+					{"new-data-resource-id", RESC_ID_KW},
+					{"new-data-size", DATA_SIZE_KW},
+					{"new-data-status", STATUS_STRING_KW},
+					{"new-data-type-name", DATA_TYPE_KW},
+					{"new-data-version", VERSION_KW}
+				});
+				// clang-format on
+
+				for (auto&& [extern_pname, intern_pname] : properties) {
```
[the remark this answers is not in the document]
korydraughn: extern -> public api param name
intern -> private irods keyword for the C API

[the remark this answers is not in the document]
alanking: Gotcha. I think my mind greatly desired/desires the "al" in external and internal :p Feel free to ignore.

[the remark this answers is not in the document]
korydraughn: Will do.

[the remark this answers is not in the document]
trel: +1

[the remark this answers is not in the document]
korydraughn: Done.

API.md
korydraughn: I not crazy about the `new-` prefix either, but for the sake of time, it felt good enough to make it clear to users what the params mean.

korydraughn: Updated documentation.

alanking (approved):
I say squash and # if happy

korydraughn:
All tests pass. Squashing and pounding.

korydraughn:
#'d

korydraughn:
Merging.

Issue #121, opened 2023-10-31
Implement OIDC state parameter


MartinFlores751:
OAuth 2.0 has an optional 'state' variable that may be passed along in the Authorization code grant, which helps combat cross-site request forgery.

We should implement this.

Additionally, we may consider the OpenID Connect 'nonce' parameter.

korydraughn:
Please close if completed.

MartinFlores751:
Forgot the additional configuration documentation, reopening until done with that.

PR #122, opened 2023-10-31
[#121] Implement OAuth state parameter

MartinFlores751:
Generate & check OAuth state value.

### DIFF
--- a/src/endpoints/authentication/impl.cpp
+++ b/src/endpoints/authentication/impl.cpp
@@ -214,6 +214,9 @@ namespace irods::http::handler
 
 			if (did_except) {
 				irods::http::globals::background_task([fn = __func__, _sess_ptr, _req = std::move(_req)] {
+					const auto timeout{irods::http::globals::oidc_configuration().at("state_timeout_in_seconds").get<int>()};
+					const auto state{irods::http::process_stash::insert(std::chrono::steady_clock::now() + std::chrono::seconds(timeout))};
+
 					body_arguments args{
 						{"client_id",
 					     irods::http::globals::oidc_configuration().at("client_id").get_ref<const std::string&>()},
@@ -221,7 +224,7 @@ namespace irods::http::handler
 						{"scope", "openid"},
 						{"redirect_uri",
 					     irods::http::globals::oidc_configuration().at("redirect_uri").get_ref<const std::string&>()},
-						{"state", "placeholder"}};
+						{"state", state}};
 
 					const auto auth_endpoint{irods::http::globals::oidc_endpoint_configuration()
 					                             .at("authorization_endpoint")
@@ -254,7 +257,29 @@ namespace irods::http::handler
 						return _sess_ptr->send(fail(status_type::bad_request));
 					}
 
-					auto is_state_valid{[](std::string_view in_state) { return in_state.compare("placeholder") == 0; }};
+					const auto is_state_valid{[](const std::string& _in_state) {
+					  const auto mapped_value{irods::http::process_stash::find(_in_state)};
+					  if (!mapped_value) {
+						return false;
+					  }
+					  
+					  const auto* expire_time{boost::any_cast<const std::chrono::steady_clock::time_point>(&*mapped_value)};
+					  if (expire_time == nullptr) {
+						return false;
+					  }
+
+					  // Ensure state is not expired...
+					  const auto is_expired{std::chrono::steady_clock::now() >= *expire_time};
+
+					  // Remove item from valid states
+					  if (!irods::http::process_stash::erase(_in_state)) {
+						// Someone else got validated first!
+						return false;
+					  }
+
+					  // Return state validity
+					  return !is_expired;
+					}};
 
 					// The state is invalid (i.e. doesn't exist, or have been used)
 					if (!is_state_valid(state_iter->second)) {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -408,13 +408,13 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
 	}
 } // load_oidc_configuration
 
-class bearer_token_eviction_manager
+class process_stash_eviction_manager
 {
     net::steady_timer timer_;
     std::chrono::seconds interval_;
 
   public:
-    bearer_token_eviction_manager(net::io_context& _io, std::chrono::seconds _eviction_check_interval)
+    process_stash_eviction_manager(net::io_context& _io, std::chrono::seconds _eviction_check_interval)
         : timer_{_io}
         , interval_{_eviction_check_interval}
     {
@@ -430,28 +430,33 @@ class bearer_token_eviction_manager
                 return;
             }
 
-            log::trace("Evicting expired bearer tokens ...");
+            log::trace("Evicting expired items...");
             irods::http::process_stash::erase_if([](const auto& _k, const auto& _v) {
-                try {
-                    const auto* p = boost::any_cast<const irods::http::authenticated_client_info>(&_v);
-                    const auto erase_token = (p && std::chrono::steady_clock::now() >= p->expires_at);
+			  // Check for client bearer token
+			  const auto* client_info{boost::any_cast<const irods::http::authenticated_client_info>(&_v)};
+			  const auto erase_token{(client_info && std::chrono::steady_clock::now() >= client_info->expires_at)};
 
-                    if (erase_token) {
-                        log::debug("Evicted bearer token [{}].", _k);
-                    }
+			  // Check for OAuth 2.0 state param
+			  const auto* expire_time{boost::any_cast<const std::chrono::steady_clock::time_point>(&_v)};
+			  const auto erase_state{(expire_time && std::chrono::steady_clock::now() >= *expire_time)};
 
-                    return erase_token;
-                }
-                catch (...) {
-                }
+			  // Determine if value need to be deleted
+			  const auto erase_value{erase_token || erase_state};
 
-                return false;
+			  if (erase_token) {
+				log::debug("Evicted bearer token [{}].", _k);
+			  }
+			  else if (erase_state) {
+				log::debug("Evicted state [{}].", _k);
+			  }
+
+			  return erase_value;
             });
 
             evict();
         });
     } // evict
-}; // class bearer_token_eviction_manager
+}; // class process_stash_eviction_manager
 
 auto main(int _argc, char* _argv[]) -> int
 {
@@ -567,7 +572,7 @@ auto main(int _argc, char* _argv[]) -> int
 
         // Launch eviction check for expired bearer tokens.
         const auto eviction_check_interval = http_server_config.at(json::json_pointer{"/authentication/eviction_check_interval_in_seconds"}).get<int>();
-        bearer_token_eviction_manager eviction_mgr{ioc, std::chrono::seconds{eviction_check_interval}};
+        process_stash_eviction_manager eviction_mgr{ioc, std::chrono::seconds{eviction_check_interval}};
 
         log::info("Server is ready.");
         ioc.run();


src/endpoints/authentication/impl.cpp
```diff
@@ -254,7 +257,30 @@ namespace irods::http::handler
...
+					  auto* expire_time{boost::any_cast<std::chrono::steady_clock::time_point>(&*mapped_value)};
+
+					  // It exists!!! Let's see what we can do...
+					  if (expire_time != nullptr) {
+						// Ensure state is not expired...
+						const auto now{std::chrono::steady_clock::now()};
+						const auto is_not_expired{now < *expire_time};
+
+						// Remove item from valid states
+						irods::http::process_stash::erase(_in_state);
```
MartinFlores751, line 275: We should check the return value of this, in the case two requests use the same state value. Only one should be valid.

src/endpoints/authentication/impl.cpp
```diff
@@ -254,7 +257,30 @@ namespace irods::http::handler
...
 					}
 
-					auto is_state_valid{[](std::string_view in_state) { return in_state.compare("placeholder") == 0; }};
+					auto is_state_valid{[](const std::string& _in_state) {
+					  // Search to see if state exists
+					  auto mapped_value{irods::http::process_stash::find(_in_state)};
+					  if (!mapped_value) {
+						return false;
+					  }
+					  auto* expire_time{boost::any_cast<std::chrono::steady_clock::time_point>(&*mapped_value)};
```
MartinFlores751, line 266: `const` these perhaps?

src/endpoints/authentication/impl.cpp
```diff
@@ -214,14 +215,16 @@ namespace irods::http::handler
 
 			if (did_except) {
 				irods::http::globals::background_task([fn = __func__, _sess_ptr, _req = std::move(_req)] {
+					const auto state{irods::http::process_stash::insert(std::chrono::steady_clock::now() + std::chrono::seconds(120))};
```
MartinFlores751, line 218: Configurable expire?

Spec doesn't say it should/shouldn't expire, but we don't want or memory to run out from garbage requests.

src/main.cpp
```diff
@@ -453,6 +454,52 @@ class bearer_token_eviction_manager
...
+			  // Not of the right type...
+			  if (expire_time == nullptr) {
+				return false;
+			  }
+
+			  const auto now{std::chrono::steady_clock::now()};
+			  const auto erase_token{now >= *expire_time};
+
+			  if (erase_token) {
+				log::debug("[{}] Evicted state=[{}].", boost::core::demangle(typeid(*this).name()), _k);
```
MartinFlores751, line 492: Fancy/different logging style, should probably change to match others.

src/main.cpp
```diff
@@ -569,6 +616,9 @@ auto main(int _argc, char* _argv[]) -> int
         const auto eviction_check_interval = http_server_config.at(json::json_pointer{"/authentication/eviction_check_interval_in_seconds"}).get<int>();
         bearer_token_eviction_manager eviction_mgr{ioc, std::chrono::seconds{eviction_check_interval}};
 
+		// TODO: Don't lean on BT timer
+		oidc_state_eviction_manager oidc_cleanup{ioc, std::chrono::seconds{eviction_check_interval}};
```
MartinFlores751, line 620: Another config option for state eviction checks?

MartinFlores751 (commented):
Words.

src/endpoints/authentication/impl.cpp
korydraughn: Do we expect the "state" value to be read/fetched within some number of milliseconds if the client is behaving?

src/endpoints/authentication/impl.cpp
```diff
@@ -254,7 +257,30 @@ namespace irods::http::handler
 						return _sess_ptr->send(fail(status_type::bad_request));
 					}
 
-					auto is_state_valid{[](std::string_view in_state) { return in_state.compare("placeholder") == 0; }};
+					auto is_state_valid{[](const std::string& _in_state) {
```
korydraughn, line 260: `const auto is_state_valid`

src/endpoints/authentication/impl.cpp
```diff
@@ -254,7 +257,30 @@ namespace irods::http::handler
...
+					  if (!mapped_value) {
+						return false;
+					  }
+					  auto* expire_time{boost::any_cast<std::chrono::steady_clock::time_point>(&*mapped_value)};
+
+					  // It exists!!! Let's see what we can do...
+					  if (expire_time != nullptr) {
+						// Ensure state is not expired...
+						const auto now{std::chrono::steady_clock::now()};
+						const auto is_not_expired{now < *expire_time};
```
korydraughn, line 272: Looks like these lines can be collapsed.

src/endpoints/authentication/impl.cpp
```diff
@@ -254,7 +257,30 @@ namespace irods::http::handler
...
+					auto is_state_valid{[](const std::string& _in_state) {
+					  // Search to see if state exists
+					  auto mapped_value{irods::http::process_stash::find(_in_state)};
+					  if (!mapped_value) {
+						return false;
+					  }
+					  auto* expire_time{boost::any_cast<std::chrono::steady_clock::time_point>(&*mapped_value)};
+
+					  // It exists!!! Let's see what we can do...
+					  if (expire_time != nullptr) {
```
korydraughn, line 269: If you inverted this check, you can return early and shift the remaining lines left by one tab character.

src/endpoints/authentication/impl.cpp
korydraughn: Yes. Especially if they don't change afterwards.

src/main.cpp
korydraughn: Tell me about this?

src/main.cpp
```diff
@@ -453,6 +454,52 @@ class bearer_token_eviction_manager
     } // evict
 }; // class bearer_token_eviction_manager
 
+class oidc_state_eviction_manager
```
korydraughn, line 457: Can this be combined with the bearer token eviction manager class? Seems like it can.

src/main.cpp
```diff
@@ -569,6 +616,9 @@ auto main(int _argc, char* _argv[]) -> int
         const auto eviction_check_interval = http_server_config.at(json::json_pointer{"/authentication/eviction_check_interval_in_seconds"}).get<int>();
         bearer_token_eviction_manager eviction_mgr{ioc, std::chrono::seconds{eviction_check_interval}};
 
+		// TODO: Don't lean on BT timer
```
korydraughn, line 619: What does BT stand for?

src/main.cpp
korydraughn: I think that depends on answers to the other questions.

src/endpoints/authentication/impl.cpp
MartinFlores751: We expect some reasonable time frame for us to redirect the user, for the user to fill in their OIDC login info, and receive the data they send back.

src/main.cpp
MartinFlores751: It's very cool, `typeid(*this).name()` gets the class name, which _might_ be mangled. `boost::core::demangle(...)` ensures we just get the class name, not a mangled version.

The full log line looks like the following:
`[2023-10-31 01:18:54.469] [P:6997] [debug] [T:7009] [oidc_state_eviction_manager] Evicted state=[5ee5b649-a9e7-46d7-ab82-7853d6b53325].`

src/main.cpp
MartinFlores751: Yes, just separated them for making the purpose of the classes clear.

AKA, didn't want to pollute the original class.

src/main.cpp
MartinFlores751: BT = bearer token. Was referring to the `bearer_token_eviction_manager`'s `eviciton_check_interval` time.

src/endpoints/authentication/impl.cpp
korydraughn: In that case, I think it needs to be configurable or the window needs to be increased.

trel: a vote for configurable...

and now... a great naming!

src/main.cpp
korydraughn: That is nice. We could also hard-code the string to avoid the work to generate that name altogether.

src/main.cpp
korydraughn: Got it.

src/main.cpp
korydraughn: I think it's fine to combine them since they are so similar. Branching would essentially be based on the pointer obtained by the any_cast.

What do you think?

korydraughn: That also means both types are handled within the same locked mutex. Without that, the two eviction managers may compete on acquiring the process stash mutex.

Thoughts?

src/main.cpp
MartinFlores751: True, but then we need to remember to update strings & classes together.
More to do if we plan to spread this out to other classes/things too.

src/main.cpp
MartinFlores751: True on competing items. Can put them together. Means we have one eviction timer, though I don't think that's much of a problem.

src/endpoints/authentication/impl.cpp
MartinFlores751: Should be able to throw in a name no problem. Can always debate on the name later too.

Ideas for names welcome!

src/endpoints/authentication/impl.cpp
MartinFlores751: `const`ing a lambda didn't cross my mind, will do 👍

src/main.cpp
korydraughn: Which strings/classes would need to be updated?
Where would we use that outside of this location?

MartinFlores751: Don't think I quite follow.
Were we planning on using this for classes in logging now?

If yes, then, if we hard-code a string in a class while logging, we need to keep that string updated for every class (now and in the future), especially if we ever copy/paste code or update the class name.

The benefit of `typeid(...).name()` is that we avoid having to keep track of these class name strings, with a presumed low cost.

Either way, if we're not going to use this style(?) of logging, it doesn't make sense to keep it here.
Might be of interest in the future though, if we care about logging details in a class.

src/endpoints/authentication/impl.cpp
MartinFlores751: Done

src/endpoints/authentication/impl.cpp
MartinFlores751: `const`ed

src/endpoints/authentication/impl.cpp
MartinFlores751: Gave it a name, can debate it in new comment/issue if wanted.

src/main.cpp
MartinFlores751: Removed line, since merged with other class.

src/endpoints/authentication/impl.cpp
MartinFlores751: Done

src/endpoints/authentication/impl.cpp
MartinFlores751: Done

src/endpoints/authentication/impl.cpp
MartinFlores751: Check inverted.

src/main.cpp
MartinFlores751: Removed class, merged with other.

src/main.cpp
MartinFlores751: Done

src/main.cpp
MartinFlores751: Class removed.

trel:
I'm seeing `state_timeout` as the named configuration setting.

Consider appending `_in_seconds`?

src/main.cpp
korydraughn: > Were we planning on using this for classes in logging now?

What do you have in mind?

> If yes, then, if we hard-code a string in a class while logging, we need to keep that string updated for every class (now and in the future), especially if we ever copy/paste code or update the class name.

Yes, a valid concern. However, for this specific case, I don't see that happening. The alternative is to do the demangling in the constructor and store it for later use.

> Either way, if we're not going to use this style(?) of logging, it doesn't make sense to keep it here.
Might be of interest in the future though, if we care about logging details in a class.

I agree.

src/main.cpp
```diff
@@ -432,20 +432,22 @@ class bearer_token_eviction_manager
...
-                    return erase_token;
-                }
-                catch (...) {
-                }
+			  // Determine if value need to be deleted
+			  const auto erase_value{erase_token || erase_state};
 
-                return false;
+			  if (erase_value) {
+				log::debug("Evicted token [{}].", _k);
```
korydraughn, line 447: This message could be adjusted based on the booleans on L444.

Being more specific here would help the admin if they ever needed to confirm eviction of something.

src/main.cpp
```diff
@@ -432,20 +432,22 @@ class bearer_token_eviction_manager
 
             log::trace("Evicting expired bearer tokens ...");
```
korydraughn, line 433: This message needs to be adjusted.

src/main.cpp
korydraughn, line 1: Don't forget to adjust the name of the eviction manager class now that it handles more than just bearer tokens.

src/main.cpp
MartinFlores751: Done

src/main.cpp
MartinFlores751: Done.

src/main.cpp
MartinFlores751: Changed name.

korydraughn (commented):
If the tests pass, squash it.

src/endpoints/authentication/impl.cpp
```diff
@@ -254,7 +258,29 @@ namespace irods::http::handler
...
+						return false;
+					  }
+					  
+					  const auto* expire_time{boost::any_cast<const std::chrono::steady_clock::time_point>(&*mapped_value)};
+					  if (expire_time == nullptr) {
+						return false;
+					  }
+
+					  // Ensure state is not expired...
+					  const auto is_not_expired{std::chrono::steady_clock::now() < *expire_time};
```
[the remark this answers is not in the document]
trel: agreed - negatives are harder to read, once you're away from the code for a while (or new).

src/main.cpp
```diff
@@ -408,13 +408,13 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
 	}
 } // load_oidc_configuration
 
-class bearer_token_eviction_manager
+class process_stash_eviction_manager
```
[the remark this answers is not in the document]
korydraughn: I don't understand. Please elaborate.

src/endpoints/authentication/impl.cpp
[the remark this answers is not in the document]
MartinFlores751: Done.

src/main.cpp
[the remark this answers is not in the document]
korydraughn: I don't think we need to worry about refactoring the eviction manager class. It probably won't see any additional changes after this PR is merged because it only does one thing. That being, walk the process stash and delete expired data.

_This is my thought after @alanking explained what he meant offline._

korydraughn:
All comments have been resolved.

@alanking Any remaining thoughts before the squash?

alanking (commented):
Seems fine to me

korydraughn:
Alrighty. @MartinFlores751 Squash it!

MartinFlores751:
One final change to remove an old mutex include. All test passed on that change.

MartinFlores751:
Squashed.

korydraughn (approved):
Pound it.

MartinFlores751:
Pounded.

PR #136, opened 2023-11-02
Multiple little tweaks

korydraughn:


src/endpoints/data_objects/impl.cpp
```diff
@@ -1605,6 +1605,14 @@ namespace
 					const json input{{"logical_path", lpath_iter->second}, {"options", options}};
 
 					auto conn = irods::get_connection(client_info.username);
+
+					if (!fs::client::is_data_object(conn, lpath_iter->second)) {
```
alanking, line 1611: Does this prevent touch from creating data objects? If so, should we check for the "no-create" option here?

alanking (commented):
I know the tests are WIP, but... still going to suggest tests :)

src/endpoints/data_objects/impl.cpp
korydraughn: Good catch! I think a better solution is to use `is_collection` instead.

The intent behind that line is to not allow references to collections.

Will change.

korydraughn: Decided it was better to test for existence and then check if the target is NOT a data object. The current code is only wrong in the case where the path doesn't exist.

korydraughn: Done. Please review and let me know if anything sticks out to you. Thanks.

korydraughn:
Added tests. Everything is passing.

alanking (commented):
Very nice. I say squash it once any open conversations are resolved.

test/test_irods_http_api.py
```diff
@@ -582,6 +579,146 @@ def test_common_operations(self):
...
+                'lpath': data_object,
+                'ppath': physical_path,
+                'resource': other_resource,
+                'as-additional-replica': 1
+            })
+            #print(r.content) # Debug
+            self.assertEqual(r.status_code, 200)
+            self.assertEqual(r.json()['irods_response']['error_code'], 0)
+
+            # Show a new data object exists with the expected replica information.
```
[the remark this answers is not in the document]

korydraughn:
Ignore clang-format. It's producing the wrong result for some odd reason.

All tests pass. Adding pounds.

alanking (approved):

korydraughn:
#'d

PR #146, opened 2023-11-07
Add support for the anonymous user account and record python dependencies for testing

korydraughn:
All tests pass with this change.

Also, the server is now detecting the absence of the anonymous user and reporting the appropriate error. I think the original behavior I was seeing was due to old build artifacts. Recompiling from scratch resolved the issue.

### DIFF
--- a/API.md
+++ b/API.md
@@ -8,13 +8,15 @@ If you discover that some topic related to the behavior of the endpoint operatio
 
 ### Scheme: Basic
 
-_Command:_
+#### Request
+
 ```bash
-curl http://localhost:<port>/irods-http-api/<version>/authenticate \
-    -X POST --user <username>:<password> 
+curl -X POST -u <username>:<password> http://localhost:<port>/irods-http-api/<version>/authenticate
 ```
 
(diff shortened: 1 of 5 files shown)


trel:
oh, well that's the best answer...

API.md
```diff
@@ -8,13 +8,15 @@ If you discover that some topic related to the behavior of the endpoint operatio
 
 ### Scheme: Basic
 
-_Command:_
+#### Request
+
 ```bash
-curl http://localhost:<port>/irods-http-api/<version>/authenticate \
-    -X POST --user <username>:<password> 
+curl -X POST -u <username>:<password> http://localhost:<port>/irods-http-api/<version>/authenticate
```
trel, line 14: just confirming...  password is plaintext here?

korydraughn: Yes.

If `-u <username>` is used instead, curl will prompt for the password without echoing the input. Mentioning that is out of scope IMO.

trel: got it.   carry on.

trel:
looks good - appears ready for squash/pound to me.

korydraughn:
#'d

trel:
mergin'.

PR #154, opened 2023-11-19
43.readme

trel:


README.md
```diff
@@ -2,6 +2,70 @@
 
 A project that presents an iRODS Zone as HTTP v1.1.
 
+## Quickstart (Running from Docker Hub Image)
+
+1. Generate a local configuration JSON file.
```
korydraughn, line 7: I don't think the numbering is needed since each step is followed by a code block and the quickstart section is very short. This also makes the instructions similar to the rest of the document.

The indentation that markdown applies also makes it less appealing visibly. Perhaps there's a way to tell markdown not to indent the numbering.

README.md
```diff
@@ -2,6 +2,70 @@
...
+2. Edit/update the template `config.json` file (point to your iRODS Zone, remove OIDC and TLS sections, etc.).
+
+```
+vim config.json
+```
+
+3. Launch the HTTP API with your customized configuration file to check for success/errors:
+
+```
+docker run --rm --name irods_http_api -v ./config.json:/config.json:ro -p 9000:9000 irods/irods_http_api
```
korydraughn, line 22: Some of these lines run off the page which forces the reader to use the horizontal scroll bar.

Consider wrapping the lines using a trailing backslash.

README.md
```diff
@@ -2,6 +2,70 @@
...
+```
+
+4. Launch in daemonized form:
+
+```
+docker run -d --rm --name irods_http_api -v ./config.json:/config.json:ro -p 9000:9000 irods/irods_http_api
```
korydraughn, line 28: Consider wrapping the lines using a trailing backslash.

README.md
korydraughn, line 22: Does this need more `-it` to view output?

Is this and the daemonized form needed? Perhaps we can leave the daemonized form as an exercise for the reader and step 4 becomes ... "In a new terminal, run the following ...".

README.md
trel: agreed.  removing numbering.

README.md
trel: will do.

README.md
trel: will do.

README.md
trel: Hmm, could be - I tested this and was able to see the 'broken' output for debugging (wrong host, bad port, etc.).

I initially only did the `-d` but couldn't see anything, so thought it would be helpful for holding/filling the `config.json` correctly.

trel:
updated - please eyeball

README.md
korydraughn: I think preferring the foreground version (i.e. no `-d`) is good for quickstart. It causes the log output to be shown immediately. Running with `-d` is a exercise for the reader IMO. Especially once they start thinking about how to deploy it in a production environment.

When you launched it without `-it`, were you able to shut it down using `CTRL-C`?

README.md
```diff
@@ -2,6 +2,78 @@
 
 A project that presents an iRODS Zone as HTTP v1.1.
 
+## Quickstart (Running from Docker Hub Image)
+
+Generate a local configuration JSON file.
+
+```
+docker run irods/irods_http_api \
+    --dump-config-template > config.json
```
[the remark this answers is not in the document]
trel: ah, good, on both counts.

README.md
trel: yes, single `Ctrl-C` exited cleanly.

trel: removed daemonized form.

korydraughn: Cool cool.

trel: resolving.

README.md
```diff
@@ -2,6 +2,68 @@
...
+docker run --rm irods/irods_http_api --dump-config-template > config.json
+```
+
+Edit/update the template `config.json` file (point to your iRODS Zone, remove OIDC and TLS sections, etc.).
+
+```
+vim config.json
+```
+
+Launch the HTTP API with your customized configuration file to check for success/errors:
```
korydraughn, line 19: The statements above don't end with `:`. Perhaps this and the one below can be changed to periods too.

Do we need to say anything about checking errors here? Seems we can just say launch the http server and that's it. If there are errors, the server will yell really loud assuming it's a startup issue.

README.md
```diff
@@ -2,6 +2,68 @@
...
+      "name": "alice",
+      "perm": "read_object",
+      "type": "groupadmin",
+      "zone": "tempZone"
+    }
+  ],
+  "registered": true,
+  "type": "collection"
+}
+```
```
korydraughn, line 61: This part feels like it should be part of a slide deck rather than a quickstart because it shows the commands along with their output.

It feels like this should be the commands followed by a section saying "If everything succeeded, you should see output similar to the following."

For example:
```
bearer_token=$(curl -X POST ...)
curl -G -H "Authorization: Bearer $bearer_token" ...
```

Thoughts?

trel: Well, I want the first-time reader to 'see' the output and get what success looks like in the first couple glances.

Hiding the token is cleaner, but ... doesn't feel as instructive.

Happy to tweak formatting to convey 'command' and then 'output' a little better.  But I think this is the right balance of 'showing' rather than 'telling'.

README.md
trel: changed two colons to periods.

leaving note about success/errors because they're never going to get it right on the first try.

trel:
pushed one more to squash.

korydraughn (approved):
Squash and pound it.

trel:
done - please merge when ready.

PR #177, opened 2023-12-19
[irods/irods#7265,irods/irods#7434] Minor reorganization + CMake TLC

SwooshyCueb:
Addresses irods/irods#7265
Addresses irods/irods#7434

I've reorganized the project to follow the conventions of the main irods repo. The individual endpoints seemed to be organized in such a way that they could be expanded into multiple source files. I have preserved this in the reorg.
Additionally, a lot of the CMake was re-written. I stuck with object libraries for everything in order to make it easier to set properties for groups of source files.

I'm not sure how to run tests for this project, but I would be surprised if this PR causes any test to fail, as there are no code changes aside from header paths.


endpoints/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_subdirectory(shared)
+
+add_subdirectory(authentication)
+add_subdirectory(collections)
+#add_subdirectory(config)
```
alanking, line 5: I didn't see config anywhere before... Is this something that's planned?

alanking (commented):
Looks fine to me - had one naive question. On that note, definitely deferring to @korydraughn and @MartinFlores751 on this one. :)

endpoints/CMakeLists.txt
SwooshyCueb: It's referenced in a couple commented out places here:
https://github.com/irods/irods_client_http_api/blob/38d67aa5ab798952450dc4429c20d82586a375e8/src/endpoints/CMakeLists.txt#L35
https://github.com/irods/irods_client_http_api/blob/38d67aa5ab798952450dc4429c20d82586a375e8/src/CMakeLists.txt#L56
And one place where it's not commented out:
https://github.com/irods/irods_client_http_api/blob/38d67aa5ab798952450dc4429c20d82586a375e8/src/endpoints/handlers.hpp#L19

korydraughn: It is a placeholder for when we decide to allow modifying iRODS configuration remotely.

korydraughn (commented):
Looks fine.

Are there other benefits to reorganizing the project structure?

SwooshyCueb:
> Are there other benefits to reorganizing the project structure?

My main goal with the reorganization was to neaten up the project tree; separating header files from source files from cmake files, putting headers in subdirectories to eliminate potential issues with generic filenames, that sort of thing.

CMakeLists.txt
```diff
@@ -94,7 +83,44 @@ else()
...
+  irods_http_api_endpoint_collections
+  #irods_http_api_endpoint_config
+  irods_http_api_endpoint_data_objects
+  irods_http_api_endpoint_information
+  irods_http_api_endpoint_query
+  irods_http_api_endpoint_resources
+  irods_http_api_endpoint_rules
+  irods_http_api_endpoint_tickets
+  irods_http_api_endpoint_users_groups
+  irods_http_api_endpoint_zones
```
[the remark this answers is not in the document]
SwooshyCueb: I did consider this and decided against it. Given how variable scoping in CMake works, this would just make the CMake really messy.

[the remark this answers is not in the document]
korydraughn: I'm a little confused.

Wouldn't introducing variables be identical to how we handle `CMAKE_IRODS_HTTP_SOURCE_DIR` above or is there more to it that I'm not seeing?

CMakeLists.txt
```diff
@@ -94,7 +83,44 @@ else()
   set(CMAKE_EXE_LINKER_FLAGS_INIT "${CMAKE_EXE_LINKER_FLAGS_INIT} -Wl,-z,defs")
 endif()
 
-add_subdirectory(src)
+set(IRODS_ENABLE_GENQUERY2 OFF CACHE BOOL "Enable GenQuery2 support. GenQuery2 package must be installed.")
+
+set(CMAKE_IRODS_HTTP_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
```
korydraughn, line 88: My question about the top-level object target names raised another question.

Does `CMAKE_` here carry any special rules? Why would we use `CMAKE_` as the prefix rather than `IRODS_`?

korydraughn:
> My main goal with the reorganization was to neaten up the project tree; separating header files from source files from cmake files, putting headers in subdirectories to eliminate potential issues with generic filenames, that sort of thing.

Makes sense.

CMakeLists.txt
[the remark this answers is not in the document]
SwooshyCueb: Oh, sorry, I think I misread/misinterpreted your original comment. I can do that.

CMakeLists.txt
SwooshyCueb: We use similar variables in the irods project that start with `CMAKE_IRODS_`. I think I did this to match `CMAKE_CURRENT_SOURCE_DIR` and `CMAKE_CURRENT_BINARY_DIR`. I'm not married to it though.

CMakeLists.txt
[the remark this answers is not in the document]
SwooshyCueb: Before I go ahead with this, under what circumstances would we want to change the target name for an object library, but not the name of the variable containing the target name?

CMakeLists.txt
korydraughn: All good. I was just wondering. I don't have any strong opinions on it at this time.

CMakeLists.txt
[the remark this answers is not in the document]
korydraughn: I'm not sure. Perhaps the names will never change. I just noticed references across files and thought maybe we should use a variable so we can change things in one place.

This isn't a requirement either _(we're likely a long way from 1.0.0)_. Just trying to get everyone to think about it in case it hadn't crossed their mind.

CMakeLists.txt
SwooshyCueb: On further reflection, and after a quick peek at the [CMake docs](https://cmake.org/cmake/help/v3.28/manual/cmake-language.7.html#variables), I think I am going to change this. I don't think we should be declaring our own `CMAKE_` variables, at least not for this case.

korydraughn: I agree.

CMakeLists.txt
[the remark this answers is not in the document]
SwooshyCueb: I don't think I'm going to make this change. I think it makes sense to have a variable name for the binary, but not for object libraries, as the variable names for object libraries would be tied too closely with the target names. If we change the target names, we'd probably end up changing the variable names too, defeating the purpose.

korydraughn:
To run the tests, you need an iRODS 4.3.1 server, the python requests module, and a valid config file for the HTTP API (see README) and a valid config file the test suite (i.e. test/config.py).

Then it's just a matter of running test_irods_http_api.py.

Testing OIDC shouldn't be necessary if the normal tests pass.

SwooshyCueb:
All normal tests pass!

korydraughn (commented):
Please squash to taste.

SwooshyCueb:
Squashed!

korydraughn:
Pound it.

SwooshyCueb:
\#'d

PR #197, opened 2024-01-05
Update OIDC configuration, use boost url

MartinFlores751:
Simplify OIDC configuration by using one 'provider_url', instead of three different parameters.

Additionally, use boost url in the parsing of OIDC urls.

### DIFF
--- a/core/include/irods/private/http_api/common.hpp
+++ b/core/include/irods/private/http_api/common.hpp
@@ -11,6 +11,7 @@
 #include <boost/beast/http/status.hpp>
 #include <boost/beast/http/message.hpp>
 #include <boost/beast/http/string_body.hpp>
+#include <boost/url/parse.hpp>
 #include <boost/uuid/uuid.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
@@ -161,6 +162,8 @@ namespace irods::http
 		request_type& _req,
 		const std::unordered_map<std::string, handler_type>& _op_table_get,
 		const std::unordered_map<std::string, handler_type>& _op_table_post) -> void;
+
+	auto get_port_from_url(boost::urls::url_view _url) -> std::optional<std::string>;
 } // namespace irods::http
 
 namespace irods
--- a/core/src/common.cpp
+++ b/core/src/common.cpp
@@ -328,6 +328,25 @@ namespace irods::http
 		log::error("{}: HTTP method not supported.", __func__);
 		return _sess_ptr->send(irods::http::fail(status_type::method_not_allowed));
 	} // operation_dispatch
+
+	auto get_port_from_url(boost::urls::url_view _url) -> std::optional<std::string>
+	{
+		if (_url.has_port()) {
+			return _url.port();
+		}
+
+		switch (_url.scheme_id()) {
+			case boost::urls::scheme::https:
+				log::debug("{}: Detected HTTPS scheme, using port 443.", __func__);
+				return "443";
+			case boost::urls::scheme::http:
+				log::debug("{}: Detected HTTP scheme, using port 80.", __func__);
+				return "80";
+			default:
+				log::error("{}: Cannot deduce port from url [{}].", __func__, _url.data());
+				return std::nullopt;
+		}
+	} // get_port_from_url
 } // namespace irods::http
 
 namespace irods
--- a/core/src/main.cpp
+++ b/core/src/main.cpp
@@ -23,6 +23,7 @@
 #include <boost/config.hpp>
 #include <boost/algorithm/string.hpp>
 #include <boost/program_options.hpp>
+#include <boost/url/parse.hpp>
 
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wunused-parameter"
@@ -225,17 +226,10 @@ constexpr auto default_jsonschema() -> std::string_view
                                 "timeout_in_seconds"
                             ]
                         }},
-                        "oidc": {{
+                        "openid_connect": {{
                             "type": "object",
                             "properties": {{
-                                "config_host": {{
-                                    "type": "string"
-                                }},
-                                "port": {{
-                                    "type": "integer",
-                                    "minimum": 1
-                                }},
-                                "well_known_uri": {{
+                                "provider_url": {{
                                     "type": "string"
                                 }},
                                 "client_id": {{
@@ -249,9 +243,7 @@ constexpr auto default_jsonschema() -> std::string_view
                                 }}
                             }},
                             "required": [
-                                "config_host",
-                                "port",
-                                "well_known_uri",
+                                "provider_url",
                                 "client_id",
                                 "redirect_uri",
                                 "irods_user_claim"
@@ -268,7 +260,7 @@ constexpr auto default_jsonschema() -> std::string_view
                         {{
                              "required": [
                                 "eviction_check_interval_in_seconds",
-                                "oidc"
+                                "openid_connect"
                              ]
                         }}
                     ]
@@ -460,10 +452,8 @@ auto print_configuration_template() -> void
                 "timeout_in_seconds": 3600
             }},
 
-            "oidc": {{
-                "config_host": "<string>",
-                "port": 8080,
-                "well_known_uri": "<string>",
+            "openid_connect": {{
+                "provider_url": "<string>",
                 "client_id": "<string>",
                 "redirect_uri": "<string>",
                 "irods_user_claim": "<string>"
@@ -696,11 +686,11 @@ auto init_irods_connection_pool(const json& _config) -> std::unique_ptr<irods::c
 		opts);
 } // init_irods_connection_pool
 
-auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpoint_config) -> void
+auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpoint_config) -> bool
 {
 	try {
 		// Setup config
-		_oi_config = _config.at(json::json_pointer{"/authentication/oidc"});
+		_oi_config = _config[json::json_pointer{"/authentication/openid_connect"}];
 		irods::http::globals::set_oidc_configuration(_oi_config);
 
 		// Consider reusing context further down main?
@@ -709,21 +699,32 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
 		beast::tcp_stream tcp_stream{io_ctx};
 
 		// Load config
-		const auto host{_oi_config.at("config_host").get_ref<const std::string&>()};
-		const auto port{std::to_string(_oi_config.at("port").get<int>())};
-		const auto temp_uri{_oi_config.at("well_known_uri").get_ref<const std::string&>()};
-		const auto uri{fmt::format("{}/.well-known/openid-configuration", temp_uri)};
+		const auto& provider{_oi_config.at("provider_url").get_ref<const std::string&>()};
+		const auto parsed_uri{boost::urls::parse_uri(provider)};
+
+		if (parsed_uri.has_error()) {
+			log::error("Error trying to parse provider_url [{}]. Please check configuration.", provider);
+			return false;
+		}
+
+		const auto url{*parsed_uri};
+		const auto path{fmt::format("{}/.well-known/openid-configuration", url.path())};
+		const auto port{irods::http::get_port_from_url(url)};
+
+		if (!port) {
+			return false;
+		}
 
 		// Resolve addr
-		const auto resolve{tcp_res.resolve(host, port)};
+		const auto resolve{tcp_res.resolve(url.host(), *port)};
 
 		// Connect and get config
 		tcp_stream.connect(resolve);
 
 		// Build Request
 		constexpr auto version_number{11};
-		beast::http::request<beast::http::string_body> req{beast::http::verb::get, uri, version_number};
-		req.set(beast::http::field::host, host);
+		beast::http::request<beast::http::string_body> req{beast::http::verb::get, path, version_number};
+		req.set(beast::http::field::host, url.host());
 		req.set(beast::http::field::user_agent, irods::http::version::server_name);
 
 		// Send request
@@ -747,7 +748,10 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
 	}
 	catch (const json::out_of_range& e) {
 		log::trace("Invalid OIDC configuration, ignoring. Reason: {}", e.what());
+		return false;
 	}
+
+	return true;
 } // load_oidc_configuration
 
 class process_stash_eviction_manager
@@ -871,8 +875,17 @@ auto main(int _argc, char* _argv[]) -> int
 		// JSON configs needs to be in main scope to last the entire duration of the program
 		nlohmann::json oi_config;
 		nlohmann::json endpoint_config;
-		load_oidc_configuration(http_server_config, oi_config, endpoint_config);
 
+		// Check if OIDC config exists, skip setup if missing.
+		if (http_server_config.contains(json::json_pointer{"/authentication/openid_connect"})) {
+			if (!load_oidc_configuration(http_server_config, oi_config, endpoint_config)) {
+				log::error("Invalid OIDC configuration, server not starting.");
+				return 1;
+			}
+		}
+		else {
+			log::info("No OIDC configuration detected, running without OIDC features.");
+		}
 		// TODO For LONG running tasks, see the following:
 		//
 		//   - https://stackoverflow.com/questions/17648725/long-running-blocking-operations-in-boost-asio-handlers
--- a/endpoints/authentication/src/main.cpp
+++ b/endpoints/authentication/src/main.cpp
@@ -19,6 +19,7 @@
 #include <boost/asio.hpp>
 #include <boost/beast.hpp>
 #include <boost/beast/http.hpp>
+#include <boost/url/parse.hpp>
 
 #include <iterator>
 #include <nlohmann/json.hpp>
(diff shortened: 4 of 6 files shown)


core/src/main.cpp
```diff
@@ -511,7 +528,9 @@ auto main(int _argc, char* _argv[]) -> int
 		// JSON configs needs to be in main scope to last the entire duration of the program
 		nlohmann::json oi_config;
 		nlohmann::json endpoint_config;
-		load_oidc_configuration(http_server_config, oi_config, endpoint_config);
+		if (!load_oidc_configuration(http_server_config, oi_config, endpoint_config)) {
```
MartinFlores751, line 531: Did we know if we wanted to fail here, or just log that OIDC couldn't be set up?

endpoints/authentication/src/main.cpp
```diff
@@ -58,37 +59,16 @@ namespace irods::http::handler
...
-		}
-
-		// Get host
-		char* host{};
-		rc = curl_url_get(endpoint, CURLUPART_HOST, &host, 0);
-		if (rc != 0) {
-			log::debug("Something happend....");
-		}
+		auto parsed_uri{boost::urls::parse_uri(token_endpoint)};
+		auto url{*parsed_uri};
```
MartinFlores751, line 63: We should probably still check the `parsed_uri`, just not sure how we want to 'fail' here...

MartinFlores751 (commented):
A few comments to think about.

core/include/irods/private/http_api/common.hpp
```diff
@@ -143,6 +144,8 @@ namespace irods::http
 	auto parse_url(const request_type& _req) -> url;
 
 	auto resolve_client_identity(const request_type& _req) -> client_identity_resolution_result;
+
+	auto get_port_from_url(boost::urls::url_view _url) -> std::optional<std::string>;
```
korydraughn, line 148: Is it recommended to copy `url_view`?

core/src/common.cpp
```diff
@@ -253,6 +253,24 @@ namespace irods::http
...
+
+	auto get_port_from_url(boost::urls::url_view _url) -> std::optional<std::string> {
+		if (_url.has_port()) {
+			return _url.port();
+		}
+
+		switch (auto scheme_type{_url.scheme_id()}; scheme_type) {
+		case boost::urls::scheme::https:
+			return "443";
+			break;
```
korydraughn, line 265: `break` can be removed.

core/src/common.cpp
```diff
@@ -253,6 +253,24 @@ namespace irods::http
...
+		case boost::urls::scheme::http:
+			return "80";
+			break;
```
korydraughn, line 268: `break` can be removed.

core/src/common.cpp
```diff
@@ -253,6 +253,24 @@ namespace irods::http
...
 	} // resolve_client_identity
+
+	auto get_port_from_url(boost::urls::url_view _url) -> std::optional<std::string> {
+		if (_url.has_port()) {
+			return _url.port();
+		}
+
+		switch (auto scheme_type{_url.scheme_id()}; scheme_type) {
+		case boost::urls::scheme::https:
+			return "443";
```
korydraughn, line 264: Let's log a message stating which port will be used. It can be a debug or trace level message.

core/src/main.cpp
```diff
@@ -363,20 +364,33 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
...
+		const auto provider{_oi_config.at("provider_url").get_ref<const std::string&>()};
+		auto parsed_uri{boost::urls::parse_uri(provider)};
+
+		if (parsed_uri.has_error()) {
+			log::error("{}: Error trying to parse 'provider_url'. Please check configuration.", __func__);
+			return false;
+		}
+
+		auto url{*parsed_uri};
+		auto host{url.host()};
```
korydraughn, line 376: Are copies intended here?

Could be marked `const`.

core/src/common.cpp
```diff
@@ -253,6 +253,24 @@ namespace irods::http
...
+			break;
+		case boost::urls::scheme::http:
+			return "80";
+			break;
+		default:
+			log::error("{}: Cannot deduce port from url.", __func__);
```
alanking, line 270: If possible, consider printing the URL in the error message as well? Perhaps it doesn't add much...

core/include/irods/private/http_api/common.hpp
MartinFlores751: Since it's a view, it holds a non-owning ref/pointer to the data. This should make it inexpensive to copy, like `string_view`.

A skim of the docs doesn't show a preference for either one.

core/src/main.cpp
```diff
@@ -363,20 +364,33 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
 		beast::tcp_stream tcp_stream{io_ctx};
 
 		// Load config
-		const auto host{_oi_config.at("config_host").get_ref<const std::string&>()};
-		const auto port{std::to_string(_oi_config.at("port").get<int>())};
-		const auto temp_uri{_oi_config.at("well_known_uri").get_ref<const std::string&>()};
-		const auto uri{fmt::format("{}/.well-known/openid-configuration", temp_uri)};
+		const auto provider{_oi_config.at("provider_url").get_ref<const std::string&>()};
```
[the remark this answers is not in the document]
MartinFlores751: Would that also do the lifetime extension in this case?

Wasn't sure so I avoided `const auto&`.

core/src/main.cpp
```diff
@@ -363,20 +364,33 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
...
+
+		auto port{irods::http::get_port_from_url(url)};
+
+		if (!port) {
+			return false;
+		}
 
 		// Resolve addr
-		const auto resolve{tcp_res.resolve(host, port)};
+		const auto resolve{tcp_res.resolve(host, *port)};
```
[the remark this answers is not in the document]
MartinFlores751: Can be.

core/src/common.cpp
MartinFlores751: Thought about that too, just wasn't sure how to get the full string out of `url_view`.

Think I could just `.data()` to get it. I'll try it out.

core/src/main.cpp
MartinFlores751: Agree on caller handling anything.

Did we have a way to say don't use OIDC?

Currently, if there's an invalid config, we just keep running and assume that the user wants to run with OIDC disabled.
Might be a new issue if we don't have a way to properly turn it off.

core/include/irods/private/http_api/common.hpp
korydraughn: Works for me.

core/src/main.cpp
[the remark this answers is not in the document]
korydraughn: I don't think so. The JSON object is read only at this point and I think `get_ref` returns a reference to the string in the global object.

Please confirm though. Check the docs for that member function.

core/src/main.cpp
trel: my vote....

if they have defined any oidc settings, and they are not good - we should just stop.   no ambiguity.  no surprises for the admin.

they can not have *any*, or they should have *all* and be good/valid.

korydraughn: In my PR, I added config validation so that should help.

What is considered an invalid config? Can the application detect that?

trel: a fair point... maybe it's not this code's job.

trel: especially if we ALWAYS run validation

of course, if they pass in a looser schema, and pass mandatory validation... getting to this code, which doesn't check... could be bad?   hmmm

MartinFlores751: Invalid could be something like an endpoint that doesn't lead to a valid OIDC config, an endpoint that doesn't use a supported scheme... Effectively if there's something wrong with `provider_url` or what it points to, I think that would fall under invalid.

MartinFlores751: > my vote....
> 
> if they have defined any oidc settings, and they are not good - we should just stop. no ambiguity. no surprises for the admin.
> 
> they can not have _any_, or they should have _all_ and be good/valid.

Would that mean if the OIDC section is missing, we skip OIDC, or we have something like an `is_enabled` do the job of that?

Thinking about how this might affect the config validation that Kory made.
That and what to check for in the HTTP code.

korydraughn: If OIDC is defined, it is enabled. If OIDC is not defined, ignore OIDC code.

I made the OIDC stanza optional in the validation commit, so we should be good.

trel: > If OIDC is defined, it is enabled. If OIDC is not defined, ignore OIDC code.

i think i agree.

at least until we hear that it's hard to hold for admins... then perhaps an 'enabled' flag could make sense.

core/src/common.cpp
```diff
@@ -253,6 +253,24 @@ namespace irods::http
...
+		if (_url.has_port()) {
+			return _url.port();
+		}
+
+		switch (auto scheme_type{_url.scheme_id()}; scheme_type) {
+		case boost::urls::scheme::https:
+			return "443";
+			break;
+		case boost::urls::scheme::http:
+			return "80";
```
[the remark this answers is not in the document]
MartinFlores751: Added a log message this, but, should we be 'louder' in the logs when using http?

Also trying to consider if it is a good idea to use this for the urls retrieved from the 'well-known openid configuration'.

core/src/main.cpp
MartinFlores751: Added logic to skip OIDC configuration if `openid_connect` is not in server config.

core/src/common.cpp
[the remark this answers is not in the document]
trel: definitely can't hurt at the 'debug' level

endpoints/authentication/src/main.cpp
MartinFlores751: Checked.

core/src/common.cpp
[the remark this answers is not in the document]
MartinFlores751: Leaving as is then.

core/src/main.cpp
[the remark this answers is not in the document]
MartinFlores751: Added `&`.

core/src/main.cpp
MartinFlores751: `const`ed any new variables I could.

core/src/main.cpp
```diff
@@ -866,8 +881,16 @@ auto main(int _argc, char* _argv[]) -> int
 		// JSON configs needs to be in main scope to last the entire duration of the program
 		nlohmann::json oi_config;
 		nlohmann::json endpoint_config;
-		load_oidc_configuration(http_server_config, oi_config, endpoint_config);
 
+		// Check if OIDC config exists, skip setup if missing.
+		if (http_server_config.contains(json::json_pointer{"/authentication/openid_connect"})) {
+			if (!load_oidc_configuration(http_server_config, oi_config, endpoint_config)) {
+				log::error("Invalid OIDC configuration, server not starting.");
+				std::exit(1);
```
korydraughn, line 889: `std::exit(1)` -> `return 1`

For consistency.

core/src/main.cpp
```diff
@@ -866,8 +881,16 @@ auto main(int _argc, char* _argv[]) -> int
 		// JSON configs needs to be in main scope to last the entire duration of the program
 		nlohmann::json oi_config;
 		nlohmann::json endpoint_config;
-		load_oidc_configuration(http_server_config, oi_config, endpoint_config);
 
+		// Check if OIDC config exists, skip setup if missing.
+		if (http_server_config.contains(json::json_pointer{"/authentication/openid_connect"})) {
+			if (!load_oidc_configuration(http_server_config, oi_config, endpoint_config)) {
```
korydraughn, line 887: We can avoid multiple lookups by using `.find`.

If the iterator is valid, pass its value to `load_oidc_configuration`.

core/src/main.cpp
```diff
@@ -691,11 +692,11 @@ auto init_irods_connection_pool(const json& _config) -> std::unique_ptr<irods::c
 		opts);
 } // init_irods_connection_pool
 
-auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpoint_config) -> void
+auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpoint_config) -> bool
```
korydraughn, line 695: Does this function need a top-level reference to `http_server` or can the `openid_connect` object be passed directly?

korydraughn (commented):
Just a few more items to resolve and we're there.

core/src/main.cpp
```diff
@@ -691,11 +692,11 @@ auto init_irods_connection_pool(const json& _config) -> std::unique_ptr<irods::c
...
 {
 	try {
 		// Setup config
-		_oi_config = _config.at(json::json_pointer{"/authentication/oidc"});
+		_oi_config = _config[json::json_pointer{"/authentication/openid_connect"}];
```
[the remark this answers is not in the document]
MartinFlores751: We check the existence of the config just before this function call, so we don't need the `.at` check.

We could always add this back for consistency(?).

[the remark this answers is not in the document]
korydraughn: Sounds good. Proceed.

core/src/main.cpp
MartinFlores751: Done.

core/src/main.cpp
MartinFlores751: Removed one parameter, as was no longer needed.

core/src/main.cpp
```diff
@@ -704,21 +704,32 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
...
 		// Load config
-		const auto host{_oi_config.at("config_host").get_ref<const std::string&>()};
-		const auto port{std::to_string(_oi_config.at("port").get<int>())};
-		const auto temp_uri{_oi_config.at("well_known_uri").get_ref<const std::string&>()};
-		const auto uri{fmt::format("{}/.well-known/openid-configuration", temp_uri)};
+		const auto& provider{_oi_config.at("provider_url").get_ref<const std::string&>()};
+		const auto parsed_uri{boost::urls::parse_uri(provider)};
+
+		if (parsed_uri.has_error()) {
+			log::error("{}: Error trying to parse provider_url [{}]. Please check configuration.", __func__, provider);
```
korydraughn, line 711: Remove leading `{}:` to match consistency of other log messages in main.cpp.

core/src/main.cpp
```diff
@@ -866,8 +880,17 @@ auto main(int _argc, char* _argv[]) -> int
 		// JSON configs needs to be in main scope to last the entire duration of the program
 		nlohmann::json oi_config;
 		nlohmann::json endpoint_config;
-		load_oidc_configuration(http_server_config, oi_config, endpoint_config);
 
+		// Check if OIDC config exists, skip setup if missing.
+		if (auto openid_config{http_server_config.find(json::json_pointer{"/authentication/openid_connect"})}; openid_config != http_server_config.end()) {
+			oi_config = *openid_config;
+			if (!load_oidc_configuration(oi_config, endpoint_config)) {
```
korydraughn, line 887: Is `oi_config` needed?

core/src/main.cpp
```diff
@@ -866,8 +880,17 @@ auto main(int _argc, char* _argv[]) -> int
 		// JSON configs needs to be in main scope to last the entire duration of the program
 		nlohmann::json oi_config;
 		nlohmann::json endpoint_config;
-		load_oidc_configuration(http_server_config, oi_config, endpoint_config);
 
+		// Check if OIDC config exists, skip setup if missing.
+		if (auto openid_config{http_server_config.find(json::json_pointer{"/authentication/openid_connect"})}; openid_config != http_server_config.end()) {
```
korydraughn, line 885: `auto` -> `const auto`?

korydraughn:
Github is reporting conflicts.

core/src/main.cpp
```diff
@@ -704,21 +704,32 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
...
+			log::error("{}: Error trying to parse provider_url [{}]. Please check configuration.", __func__, provider);
+			return false;
+		}
+
+		const auto url{*parsed_uri};
+		const auto path{fmt::format("{}/.well-known/openid-configuration", url.path())};
+		const auto port{irods::http::get_port_from_url(url)};
+
+		if (!port) {
+			return false;
```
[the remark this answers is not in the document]
MartinFlores751: We could log here, but I feel like we log enough in `get_port_from_url(...)` and the calling function already.

Can always put a brief message if desired, though.

core/src/main.cpp
MartinFlores751: Reverted since `.find(...)` does not work with `json_pointer`.

core/src/main.cpp
MartinFlores751: If we're talking about the json object, I believe so. Could be wrong, though.

core/src/main.cpp
MartinFlores751: Done.

MartinFlores751:
Note: Only one test fails.
`test_server_does_not_crash_when_incorrect_http_method_is_used_to_authenticate - AssertionError: 200 != 405`

Assuming this is because the test expects the OIDC config to not be present when running?
Rebasing & retesting shortly.

korydraughn:
Looks like some changes were reverted?

MartinFlores751:
> Looks like some changes were reverted?

Yes, since the using json `.find()` does not work with `json_pointer`, I reverted back to using `.contains(...)`

MartinFlores751:
> Note: Only one test fails. `test_server_does_not_crash_when_incorrect_http_method_is_used_to_authenticate - AssertionError: 200 != 405`
> 
> Assuming this is because the test expects the OIDC config to not be present when running? Rebasing & retesting shortly.

In addition, the following two test do not pass:

`test_only_a_rodsadmin_can_change_the_type_of_a_user - AssertionError: 200 != 400`
`test_rodsusers_cannot_change_passwords - AssertionError: 200 != 400`

I don't believe any of the changes should modify this, assume it has to do with test configuration as well?

core/src/main.cpp
```diff
@@ -225,17 +226,10 @@ constexpr auto default_jsonschema() -> std::string_view
...
                             "type": "object",
                             "properties": {{
-                                "config_host": {{
-                                    "type": "string"
-                                }},
-                                "port": {{
-                                    "type": "integer",
-                                    "minimum": 1
-                                }},
-                                "well_known_uri": {{
```
korydraughn, line 238: So these are no longer needed?

core/src/main.cpp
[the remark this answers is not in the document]
korydraughn: If it's already logged, then we're good.

korydraughn (commented):
Very nice.

Are the existing OIDC tests good? Do we need any new tests?

korydraughn:
Please run clang-format and investigate why the tests fail.

core/src/main.cpp
MartinFlores751: They are all replaced by `provider_url`.

core/src/common.cpp
```diff
@@ -328,6 +328,25 @@ namespace irods::http
...
 		return _sess_ptr->send(irods::http::fail(status_type::method_not_allowed));
 	} // operation_dispatch
+
+	auto get_port_from_url(boost::urls::url_view _url) -> std::optional<std::string>
+	{
+		if (_url.has_port()) {
+			return _url.port();
+		}
+
+		switch (auto scheme_type{_url.scheme_id()}; scheme_type) {
```
korydraughn, line 338: Could just be `_url.scheme_id()`?

core/src/main.cpp
korydraughn: This has been resolved.

korydraughn:
The failures are real. I was testing against unreleased iRODS server code. There are 3 failing tests and non of them are related to your work.

I'll open an issue for the failures and get them resolved.

Please resolve any remaining review comments and squash to taste. Leave off the pounds.

MartinFlores751:
Squashed.

korydraughn:
Do we have tests for the OIDC logic?

MartinFlores751:
> Do we have tests for the OIDC logic?

I'm not sure how to make an OpenID Provider give out data to make us go down the error logic path.
I think we might have to create our own fake provider to test out those paths.

For testing the OIDC init error handling, if we had a setup to test different HTTP API server configurations, I think we could test that part too.

korydraughn:
Ok. Please verify the existing tests pass (ignoring the ones in PR #221) and attach an issue number to the commit.

Leave off pounds.

MartinFlores751:
Rebase, testing.

MartinFlores751:
Got one failed test:
`test_server_does_not_crash_when_incorrect_http_method_is_used_to_authenticate - AssertionError: 200 != 405`

korydraughn:
Which line/assertion is failing?

MartinFlores751:
Error snippet:

```
...

    def test_server_does_not_crash_when_incorrect_http_method_is_used_to_authenticate(self):
        # Try to authenticate using the HTTP GET method.
        r = requests.get(self.url_endpoint, auth=('rods', 'rods'))
        self.logger.debug(r.content)
>       self.assertEqual(r.status_code, 405)
E       AssertionError: 200 != 405

test_irods_http_api.py:177: AssertionError

...
```

korydraughn:
Once #218 is merged, please rebase and confirm all tests pass.

MartinFlores751:
All pass.

korydraughn:
Attach appropriate issue numbers to the commit and we'll eyeball it one last time.

Leave off the pounds.

korydraughn (approved):
Pound it!

MartinFlores751:
Pounded.

korydraughn:
Merging.

Issue #166, opened 2023-12-16
Add inherit boolean in set_permission operation for collection endpoint


mstfdkmn:
## Feature
Seems that there is no way to enable or disable inheritance on a given collection.

PR #209, opened 2024-01-12
[#166] Implement set_inheritance operation for /collections endpoint

korydraughn:
All tests pass.

No OIDC tests were run for this.

### DIFF
--- a/API.md
+++ b/API.md
@@ -259,6 +259,36 @@ If an HTTP status code of 200 is returned, the body of the response will contain
 
 If there was an error, expect an HTTP status code in either the 4XX or 5XX range.
 
+### set_inheritance
+
+Enable or disable inheritance on a collection.
+
+#### Request
+
+```bash
+curl http://localhost:<port>/irods-http-api/<version>/collections \
+    -H 'Authorization: Bearer <token>' \
+    --data-urlencode 'op=set_inheritance' \
+    --data-urlencode 'lpath=<string>' \
+    --data-urlencode 'enable=<integer>' \ # 0 or 1.
+    --data-urlencode 'admin=<integer>' # 0 or 1. Defaults to 0. Optional.
+```
+
+#### Response
+
+If an HTTP status code of 200 is returned, the body of the response will contain JSON. Its structure is shown below.
+
+```js
+{
+    "irods_response": {
+        "status_code": 0
+        "status_message": "string" // Optional
+    }
+}
+```
+
+If there was an error, expect an HTTP status code in either the 4XX or 5XX range.
+
 ### modify_permissions
 
 Adjust permissions for multiple users and groups on a collection atomically.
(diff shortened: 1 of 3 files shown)


endpoints/collections/src/main.cpp
```diff
@@ -689,6 +690,95 @@ namespace
...
+						res,
+						http::status::bad_request,
+						json{{"irods_response", {{"status_code", NOT_A_COLLECTION}}}}.dump()));
+				}
+
+				const auto enable_iter = _args.find("enable");
+				if (enable_iter == std::end(_args)) {
+					log::error("{}: Missing [enable] parameter.", fn);
+					return _sess_ptr->send(irods::http::fail(res, http::status::bad_request));
+				}
```
korydraughn, line 734: Move this above the call to get_connection.

API.md
```diff
@@ -259,6 +259,36 @@ If an HTTP status code of 200 is returned, the body of the response will contain
 
 If there was an error, expect an HTTP status code in either the 4XX or 5XX range.
 
+### set_inheritance
```
trel, line 262: can we 'see' the inheritance setting somewhere already?   in the stat information perhaps?

trel: oh, it's in the stat.   thanks tests!

endpoints/collections/src/main.cpp
```diff
@@ -689,6 +690,95 @@ namespace
...
+					log::error("{}: {}", fn, e.what());
+					// clang-format off
+					res.body() = json{
+						{"irods_response", {
+							{"status_code", e.code().value()},
+							{"status_message", e.what()}
+						}}
+					}.dump();
+					// clang-format on
+				}
```
alanking, line 758: This looks functionally identical to the `filesystem_error` catch just below this (L760-L764). Could we catch the `filesystem_error`s in one spot, or differentiate the error messages in some way?

alanking (commented):
Yep, seems good.

I know it's planned for some time in the future, but I will nevertheless suggest adding some tests for values other than the valid ones. :)

endpoints/collections/src/main.cpp
korydraughn: That catch block is a copy-paste situation. I must have been tired or something to let that through lol.

Will remove it.

korydraughn:
I was about to add a test for invalid values, but then ran into the question of whether values other than 0 (for disable) should be considered invalid.

We can of course check for 0 or 1 and report an error, and that's good. However, that does imply additional checking of inputs, which may not be necessary.

I can see this going either way. I'm going to open an issue for this since it applies to many operations.

I'm going to hold on adding tests for invalid values until more users appear and have opinions. Things can change drastically depending on what the community learns/says and I don't want to over commit to anything just yet.

endpoints/collections/src/main.cpp
korydraughn: Done.

korydraughn:
Created #213.

korydraughn:
Any last comments before I squash/pound?

trel:
squash em

korydraughn:
Squashed. Not sure if you meant to pound it too.

korydraughn:
Adding pound. All tests pass.

If something needs to be changed, we can address it in a separate PR.

korydraughn:
#'d

Issue #185, opened 2023-12-20
Trying to read non existent object doesn't return an error


mstfdkmn:
### iRODS Version, OS and Version
4.3.1, almalinux8

### What did you try to do?
Tried to read non-existing object. Tried to read an object that the user doesn't have access.

### Expected behavior
Expected to see `-358000` or `-818000` in irods response.

### Observed behavior (including steps to reproduce, if applicable)

```
[mustafa@dev-u0137480 ~]$ curl -H "Authorization: Bearer $token" 'http://localhost:9000/irods-http-api/0.1.0/data-objects' --data-urlencode 'op=read' --data-urlencode 'lpath=/u0137480/home/mustafa/myfile1.txt' --data-urlencode 'count=6' -G
[mustafa@dev-u0137480 ~]$

```

trel:
What does this show with `-v`?    a 404?

korydraughn:
The HTTP API uses dstream to read/write data objects. The dstream library does not expose iRODS error codes. It only allows you to check the state of the stream object (i.e. good, bad, fail, etc). This matches the behavior of `std::fstream` which is provided by the C++ standard.

So, we can't give the exact reason for why the read failed (yet).

Our only options are ...
- Use the C API directly
- Return a different HTTP status code
- Make dstream capture error codes in iRODS 4.3.2

korydraughn:
With #184, we made the read operation simpler. That change also resolved this issue (a little).

Notice the HTTP status code returned in the example.
```
$ curl 'http://localhost:9000/irods-http-api/0.1.0/data-objects' \
   -H "Authorization: Bearer $token" \
  --data-urlencode 'op=read' \
  --data-urlencode 'lpath=/tempZone/home/rods/doesnotexist' -Gv                                                                                                                                                                      

*   Trying 127.0.0.1:9000...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9000 (#0)
> GET /irods-http-api/0.1.0/data-objects?op=read&lpath=%2FtempZone%2Fhome%2Frods%2Fdoesnotexist HTTP/1.1
> Host: localhost:9000
> User-Agent: curl/7.68.0
> Accept: */*
> Authorization: Bearer ce843927-acba-47ca-a81e-9276ddc4710a
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
[...]

PR #211, opened 2024-01-12
[#185] Return appropriate HTTP status code when reading non-existent data object

korydraughn:


test/test_irods_http_api.py
```diff
@@ -1565,6 +1565,91 @@ def test_remove_operation_supports_unregistering_all_replicas(self):
             })
             logging.debug(r.content)
 
+    def test_attempting_to_read_non_existent_data_object_results_in_an_error(self):
```
trel, line 1568: please add another similar test for a file that a user does not have permission to 'get'.

it does exist, but they aren't allowed to see it.   still should be 404.   don't want to leak to them info with a 403.

trel: reading is fundamental, you did this...  exactly one test below.    carry on.  thanks.

alanking (approved):
Seems good to me.

trel:
yes, please # it when ready.

korydraughn:
#'d

Issue #165, opened 2023-12-15
authenticate endpoint doesnt handle wrong request method (GET)


mstfdkmn:
### iRODS Version, OS and Version
4.3.1, almalinux8

### What did you try to do?
Accidentally I called `GET` against the `/authenticate` endpoint.

### Expected behavior
Nothing. Expected the server is running normal and the client should encounter nothing.

### Observed behavior (including steps to reproduce, if applicable)

When the client called this,
```
[irods@dev-u0137480 go-http-client]$ curl -X GET -u rods:rods http://localhost:9000/irods-http-api/0.1.0/authenticate
curl: (52) Empty reply from server
```

The server crashed with:
```
[2023-12-15 13:50:58.641] [P:2326977] [error] [T:2327026] parse_url: curl_url_get(CURLUPART_QUERY) error: 16
Segmentation fault (core dumped)
```
The server is configured with the basic authentication schema. Might be related to #139?

trel:
Yuck.   We'll definitely shore that up.   Thanks.

trel:
Yes, looks extremely related to #139.   Need more bumpers.

Issue #139, opened 2023-11-04
Server segfaults on `parse_url`


korydraughn:
## Bug Report
I attempted to reach the /authenticate endpoint by typing the URL in Google Chrome and before I got the prompt, the server crashed with the following.
```
[2023-11-04 08:12:16.669] [P:2566900] [error] [T:2566985] parse_url: curl_url_get(CURLUPART_QUERY) error: 16
Segmentation fault (core dumped)
```

Below is what I typed in the browser.
```
http://localhost:9000/irods-http-api/0.1.0/authenticate
```

korydraughn:
This was caused by my config not having appropriate OIDC settings defined. The implementation currently replies on a GET request as part of the OIDC flow.

If we want to allow users to authenticate using Basic auth in a browser, we'll need to adjust how the HTTP API handles OIDC. Perhaps an op param can be introduced or a separate endpoint.

trel:
So we definitely need to put bumpers on that use case (missing config items).

A separate endpoint might be easier / separate the concerns.

korydraughn:
Supporting authentication using the following syntax is not a requirement of the HTTP API.
```
http://<username>:<password>@localhost:9000/irods-http-api/0.1.0/authenticate
```

For that reason, this issue has been resolved since attempting to do what's described in this issue no longer results in the server crashing.

korydraughn:
Need to add tests for this.

PR #212, opened 2024-01-12
[#139,#165] Return 405 when HTTP GET is used for basic authentication

korydraughn:

### DIFF
--- a/endpoints/authentication/src/main.cpp
+++ b/endpoints/authentication/src/main.cpp
@@ -206,6 +206,11 @@ namespace irods::http::handler
 			static const auto oidc_stanza_exists = irods::http::globals::configuration().contains(
 				nlohmann::json::json_pointer{"/http_server/authentication/oidc"});
 
+			if (!oidc_stanza_exists) {
+				log::error("{}: HTTP GET method cannot be used for Basic authentication.", __func__);
+				return _sess_ptr->send(fail(status_type::method_not_allowed));
+			}
+
 			url url;
 			bool did_except{false};
 			try {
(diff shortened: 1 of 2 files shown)


endpoints/authentication/src/main.cpp
```diff
@@ -396,6 +396,10 @@ namespace irods::http::handler
 					return _sess_ptr->send(std::move(res_rep));
 				});
 			}
+			else {
```
trel, line 399: indentation not ... quite right for the put above?   

consider putting this case at the top, since it's real short and easy to see scope?

trel (commented):
looks nice

endpoints/authentication/src/main.cpp
alanking: Hmm... could be...
```c++
if (!oidc_stanza_exists) {
    // this code
}
else if (did_except) {
    // the first block
}
else {
    // the middle block / existing else-if
}
```

I had a similar thought but wasn't sure if it would be worth doing.

korydraughn: > indentation not ... quite right for the put above?

I'm not sure what you're referring to for "for the put above?". Also, clang-format hasn't reported an issue. Does the alignment look off to you?

> consider putting this case at the top, since it's real short and easy to see scope?

Will swizzle.

trel: this is for the 'get' case... so i was referring to the 'post' above, sorry, not 'put'.

korydraughn: I thought about reordering things in the PR before this, but it was easy to do the bare minimum.

This whole source file is due for a refactor.

korydraughn: Updated.

alanking (approved):
I think it looks good. Deferring to @trel

trel:
yes, much easier to reason about / read.   thanks.

korydraughn:
#'d

PR #215, opened 2024-01-14
Documentation, HTTP status codes, and Test tweaks

korydraughn:
All tests pass.


API.md
```diff
@@ -1200,13 +1202,15 @@ curl http://localhost:<port>/irods-http-api/<version>/query \
...
-    --data-urlencode 'offset=<integer>' \ # Number of rows to skip. Defaults to 0.
-    --data-urlencode 'count=<integer>' \ # Number of rows to return. Default set by administrator.
+    --data-urlencode 'args=<string>' \ # List of arguments. Optional.
+    --data-urlencode 'args-delimiter=<string>' \ # Delimiter used to separate arguments. Defaults to comma (,). Optional.
+    --data-urlencode 'offset=<integer>' \ # Number of rows to skip. Defaults to 0. Optional.
+    --data-urlencode 'count=<integer>' \ # Number of rows to return. Default set by administrator. Optional.
     -G
 ```
 
+`count` will be clamped to the range [1, _N_] where _N_ is a value defined by the administrator of the server.
```
alanking, line 1212: I like the clarification. Is "the server" referring to the iRODS server or the HTTP API server? Would it be helpful at all to explicitly reference the configuration here?

alanking (commented):
Everything looks good to me. Had one suggestion / question.

API.md
korydraughn: **the server** refers to the HTTP API server. Can change if **the server** feels too loose. I went with that because the documentation is specific to the HTTP API.

> Would it be helpful at all to explicitly reference the configuration here?

I originally included that, but wasn't sure whether it was important for clients to know that. I can see it going either way.

Looking at this again, it makes sense to mention it here because clients can hit the /info endpoint to fetch _N_. It also makes sense to add a note about the value being discoverable via the /info endpoint.

alanking: I think... if "the server" is the way this is documented elsewhere and it's consistent, then this is fine as long as it's understood/declared/etc. that it's referring to the HTTP API server, not the iRODS server.

Even though it makes documentation upkeep a bit more of a hassle, I'm more and more over time liking the idea of being a little more explicit with configuration values and where they're coming from. That's just my thought from seeing at least a dozen issues in iRODS requesting documentation about certain configurations values and what can/cannot be done with them.

trel: can we just change 'the server' to 'the API'?

korydraughn: @alanking I totally agree. **the server** was also me attempting to see how it feels to not type out iRODS HTTP API everywhere, but being explicit is never a bad thing.

@trel **the API** works too, but if we go with that, I feel we should just go ahead and include HTTP too. Then there's no room for confusion.

I'll change to **the HTTP API**. We can adjust it later if needed.

korydraughn:
Wording has been updated. We can tweak as much as needed leading up to 1.0.0.

Adding pounds.

Issue #224, opened 2024-01-22
`test_server_does_not_crash_when_incorrect_http_method_is_used_to_authenticate` fails if the server config contains the OIDC configuration stanza


korydraughn:
The test needs to be skipped if the OIDC config is defined.

PR #229, opened 2024-01-24
[#224] Skip Basic authentication test if OIDC is enabled

korydraughn:
All tests pass (including OIDC tests).


alanking (approved):
Just confirming that the "incorrect results" in the now-skipped test are not "the server crashes" :)

I think that all of this checks out

korydraughn:
That's correct. No crashing. The status codes change depending on whether OIDC is defined.

korydraughn:
Updated the commit message and included more words in the test to help others.

#'d

alanking (approved):

Issue #249, opened 2024-02-08
Add missing required configuration options for TLS communication with iRODS server


korydraughn:
## Bug Report

The following options are required for proper TLS communication between the HTTP API and the iRODS server.
- irods_client_server_negotiation
- irods_encryption_algorithm
- irods_encryption_key_size
- irods_encryption_num_hash_rounds
- irods_encryption_salt_size

See https://groups.google.com/g/irod-chat/c/xOmm2BIwezM/m/myDK12grAAAJ for a workaround.

_Originally reported at https://groups.google.com/g/irod-chat/c/xOmm2BIwezM._

PR #250, opened 2024-02-08
[#249] Add missing TLS config options

korydraughn:


core/src/main.cpp
```diff
@@ -495,11 +515,16 @@ auto print_configuration_template() -> void
...
             "ca_certificate_file": "<string>",
             "certificate_chain_file": "<string>",
             "dh_params_file": "<string>",
-            "verify_server": "<string>"
+            "verify_server": "cert",
+            "client_server_negotiation": "request_server_negotiation",
+            "encryption_algorithm": "AES-256-CBC",
+            "encryption_key_size": 32,
+            "encryption_hash_rounds": 16,
+            "encryption_salt_size": 8
```
alanking, line 527: Are these required when the `client_server_policy` is `CS_NEG_REFUSE`?

Just wondering in light of the default/suggested values being used in combination here.

alanking (commented):
Seems good to me. I had one question for my own understanding, not a blocker by any means.

core/src/main.cpp
```diff
@@ -651,21 +676,38 @@ auto set_log_level(const json& _config) -> void
...
+	set_env_var_string("/irods_client/tls/client_server_policy", irods::KW_CFG_IRODS_CLIENT_SERVER_POLICY, "CS_NEG_REFUSE");
+	set_env_var_string("/irods_client/tls/ca_certificate_file", irods::KW_CFG_IRODS_SSL_CA_CERTIFICATE_FILE);
+	set_env_var_string("/irods_client/tls/certificate_chain_file", irods::KW_CFG_IRODS_SSL_CERTIFICATE_CHAIN_FILE);
+	set_env_var_string("/irods_client/tls/dh_params_file", irods::KW_CFG_IRODS_SSL_DH_PARAMS_FILE);
+	set_env_var_string("/irods_client/tls/verify_server", irods::KW_CFG_IRODS_SSL_VERIFY_SERVER, "cert");
+	set_env_var_string("/irods_client/tls/client_server_negotiation", irods::KW_CFG_IRODS_CLIENT_SERVER_NEGOTIATION, "request_server_negotiation");
+	set_env_var_string("/irods_client/tls/encryption_algorithm", irods::KW_CFG_IRODS_ENCRYPTION_ALGORITHM, "AES-256-CBC");
+	set_env_var_int("/irods_client/tls/encryption_key_size", irods::KW_CFG_IRODS_ENCRYPTION_KEY_SIZE, 32);
+	set_env_var_int("/irods_client/tls/encryption_hash_rounds", irods::KW_CFG_IRODS_ENCRYPTION_NUM_HASH_ROUNDS, 16);
+	set_env_var_int("/irods_client/tls/encryption_salt_size", irods::KW_CFG_IRODS_ENCRYPTION_SALT_SIZE, 8);
```
MartinFlores751, line 709: Consider having `/irods_client/tls` being a common json pointer and appending as needed.

Might reduce typos if we change this again later.

core/src/main.cpp
```diff
@@ -651,21 +676,38 @@ auto set_log_level(const json& _config) -> void
 
 auto init_tls(const json& _config) -> void
 {
-	const auto set_env_var = [&](const auto& _json_ptr_path, const char* _env_var, const char* _default_value = "") {
-		using json_ptr = json::json_pointer;
+	const auto set_env_var_string =
+		[&](const auto& _json_ptr_path, const char* _env_var, const char* _default_value = "") {
```
MartinFlores751, line 680: Just to be sure, the capture only takes `_config` right?

core/src/main.cpp
trel: they are not required if `CS_NEG_REFUSE` is used.

core/src/main.cpp
```diff
@@ -651,21 +676,38 @@ auto set_log_level(const json& _config) -> void
...
+				setenv(env_var_upper.c_str(), v.c_str(), 1); // NOLINT(concurrency-mt-unsafe)
+			}
+		};
 
-		if (const auto v = _config.value(json_ptr{_json_ptr_path}, _default_value); !v.empty()) {
-			const auto env_var_upper = boost::to_upper_copy<std::string>(_env_var);
-			log::trace("Setting environment variable [{}] to [{}].", env_var_upper, v);
-			setenv(env_var_upper.c_str(), v.c_str(), 1); // NOLINT(concurrency-mt-unsafe)
-		}
+	const auto set_env_var_int = [&](const auto& _json_ptr_path, const char* _env_var, int _default_value) {
```
MartinFlores751, line 690: Same as other capture comment:

Just to be sure, the capture only takes `_config` right?

core/src/main.cpp
```diff
@@ -360,13 +360,33 @@ constexpr auto default_jsonschema() -> std::string_view
...
                         "client_server_policy",
                         "ca_certificate_file",
                         "dh_params_file",
-                        "verify_server"
+                        "verify_server",
+                        "client_server_negotiation",
+                        "encryption_algorithm",
+                        "encryption_key_size",
+                        "encryption_hash_rounds",
+                        "encryption_salt_size"
```
trel, line 389: like @alanking said, most of these are only required if client_server_policy is not `CS_NEG_REFUSE`.

MartinFlores751 (commented):
Some small comments on the code bits.

core/src/main.cpp
korydraughn: @trel is correct.

core/src/main.cpp
korydraughn: Correct. They are only required when the `tls` property is defined. That forces the admin to set those values accordingly.

core/src/main.cpp
korydraughn: Correct.

core/src/main.cpp
korydraughn: Good catch.

Will lock that down.

core/src/main.cpp
trel: so... do we want to require them here in the schema?    doesn't seem quite right.

but like always, it's a balance between requiring, offering default examples, and the code handling a missing value and using a hard-coded default...

core/src/main.cpp
korydraughn: Well, the values of the properties are ignored by the iRODS libraries when CS_NEG_REFUSE is set.

All of the properties are required, but that only matters if the `tls` property exists in the config file.

core/src/main.cpp
```diff
@@ -676,37 +676,51 @@ auto set_log_level(const json& _config) -> void
...
-			if (const auto v = _config.value(json_ptr{_json_ptr_path}, _default_value); !v.empty()) {
+	// The iRODS connection libraries do not provide a clean way for developers to easily
+	// configure TLS without relying on an irods_environment.json file. All is not lost
+	// though. Turns out the iRODS libraries do inspect environment variables. This gives
+	// the HTTP API a way to hook into the connection logic and support TLS through its
+	// own configuration file. Hence, the following environment-based lambda functions.
+
+	const auto set_env_string =
+		[&](const auto& _tls_prop, const char* _env_var, const char* _default_value = "") {
+			const auto element_path = fmt::format("/irods_client/tls/{}", _tls_prop);
```
trel, line 687: this hard-coded irods_client/tls means that **other** environment variables are not possible...

should we pass two parameters to set_env_string and set_env_int?

no real use case yet - just thinking out loud.    ignore with confidence...

core/src/main.cpp
korydraughn: If an admin doesn't plan on enabling TLS, they don't need to include the `tls` stanza at all. Once they include that property, the schema requires them to define all of the other properties.

If we relax that, it's easier for admins to get the TLS settings incorrect.

There's also the possibility of making the schema more complicated by changing the validation based on the value of `client_server_negotiation`. I don't think making the schema more complicated is worth it at this time.

Side note: Perhaps `dh_params_file` should be optional and the chain file property should be required. Hmm.

core/src/main.cpp

core/src/main.cpp
trel: i think that logic is okay/good.   yes, if they have a tls section, they should have all of it.  that's fine.

regarding dh_params... let's require it as good practice for now.  loosen if we get pushback.    required value can always be empty string...

core/src/main.cpp
trel: agreed.

core/src/main.cpp
korydraughn: Updated.

Still deciding whether this change is worth it.

core/src/main.cpp
korydraughn: What are your thoughts on the chain file property becoming required?

Asking based on this: https://docs.irods.org/4.3.1/plugins/pluggable_authentication/#server-ssl-setup

See the first sentence of the section **Create the certificate chain file**.

trel: chain file is for the irods server...

compare with the new password booth (just a client)
https://github.com/irods/irods_client_password_booth/blob/main/app.config.template

trel: that suggests no dh_params either

korydraughn: Cool cool. Will investigate the removal of the dh_params and chain file properties and report what I find.

core/src/main.cpp
korydraughn: @alanking Does that answer your question?

Just to be even more clear, I'm guessing the client just sends whatever is defined. It's the iRODS server that decides whether it needs to worry about the other properties.

alanking: Oh, yes, it does. I think I was curious about the options being there while `CS_NEG_REFUSE` is the value we provide by default. I think it's well understood and the folks who care will know what to do because they need to have SSL/TLS enabled on the iRODS server to use it anyways, like you guys said. Resolving

core/src/main.cpp
korydraughn: Removed properties.

alanking (commented):
Seems good to me. Is there anything left to do? Maybe ready for squash?

korydraughn:
This is ready for squashing. Will handle that soon.

alanking (approved):
\# it

Issue #235, opened 2024-01-25
OIDC confidential access type and Authorization Code Grant


ll4strw:
## OIDC confidential access type and Authorization Code Grant
Hello, I am testing the http api with keycloak. Is there a conf parameter that I can set in my `config.json` such that I can authenticate against an OpenID client with a pair (client_id, client_secret) as credentials? Is this a supported feature? Also, with the following conf in place 
```
"openid_connect": {
                "provider_url": "http://localhost:8080/realms/iRODS/.well-known/openid-configuration",
                "client_id": "irods",
                "redirect_uri": "http://localhost:9000/irods-http-api/0.1.0/authenticate",
                "state_timeout_in_seconds": 600,
                "irods_user_claim": "preferred_username"
            }

```
and a public (no secret) oicd client, if I try the following 

```
curl http://localhost:9000/irods-http-api/0.1.0/authenticate -v
*   Trying 127.0.0.1:9000...
* Connected to localhost (127.0.0.1) port 9000 (#0)
> GET /irods-http-api/0.1.0/authenticate HTTP/1.1
> Host: localhost:9000
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server

``` 
I get the following error in the http api logs

```
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Header: (Host, localhost:9000)
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Header: (User-Agent, curl/7.81.0)
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Header: (Accept, */*)
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Method: GET
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Version: 11
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Target: /irods-http-api/0.1.0/authenticate
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Keep Alive: true
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Has Content Length: false
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Chunked: false
[2024-01-25 13:44:48.402] [P:1] [debug] [T:20] on_read: Needs EOF: false
[...]

Issue #155, opened 2023-11-20
Consider having the HTTP API be an OAuth resource server instead of a OAuth client


MartinFlores751:
Use [OAuth's RFC Figure 1](https://datatracker.ietf.org/doc/html/rfc6749#section-1.2) for a simple visual reference.

Currently, the HTTP API is treated as on OAuth client. This requires the HTTP API to handle multiple OAuth flows, token refresh, and additional complexities in handling various OAuth/OIDC items. (This isn't to say every feature is currently handled, rather, if we continue on, we would likely need to support these additional features.)

By treating the HTTP API as a resource server, the HTTP API may no longer need to handle all of these cases.
It simply accepts OAuth access tokens, verifies them, and verifies there is a mapping between the OAuth user and an iRODS user.
Additionally, grants such as the authorization code grant could be implemented in a much more user-friendly way, compared to how they are currently implemented.
This may also allow the HTTP API to use the OAuth access tokens directly instead of converting it into an alternative 'internal HTTP API token', leaning more on the OAuth server to determine when the token expires.

However, this would mean clients using the HTTP API wishing to support OAuth would need to handle all the desired grants themselves.
Additionally, by no longer being a client, there may be some complications in the mapping of users. Some work needs to be done to verify whether this is true, however, given the current usage of OAuth/OIDC in the HTTP API, it is unlikely to cause issues.

MartinFlores751:
Glad to see you're interested.

>  I am just wondering whether your proposal is being considered for the future or there are no concrete plans at the moment.

I am currently working on implementing this feature. When it is released, you should be able to choose between running the HTTP API in 'OAuth client' mode, or 'OAuth protected resource server' mode.

>  I am thinking of, for instance, an iRODS web app which could retrieve directly an access token from an oidc provider and use that in the backend for future API calls. Single sign-on would also work.

When it releases, you should be able to do this. The sign-in experience should be better, too.

MartinFlores751:
Some notes on possible limitations/issues:

There may be some issues getting this to work with Google's OpenID Provider.

``` mermaid
  sequenceDiagram
      autonumber

      actor HTTP API Consumer
      participant OpenID Provider
      participant iRODS HTTP API

      %% User Authentication
      HTTP API Consumer ->> OpenID Provider: POST /<OpenID Authentication>
      OpenID Provider -->> HTTP API Consumer: 302 FOUND /authenticate (payload)

      %% Exchange temp token for OIDC token
[...]

PR #252, opened 2024-02-12
[#155,#235,#108] Make HTTP API a Protected Resource, Add support of Confidential Client, User Mapping

MartinFlores751:
## Summary
The primary goal of this PR is to allow the HTTP API to receive access tokens on all of the iRODS HTTP API endpoints. We do this while moving away from the currently used HTTP API tokens in OAuth related authorization. A major reason for moving towards OAuth tokens is for better integration with existing providers, as well as improved compliance to desired token properties (e.g. token expiration, revocation).

As part of this implementation, confidential clients and user mapping needed to be implemented. The implementation of confidential clients should allow for the HTTP API to contact the OpenID Provider's introspection endpoint, among other benefits. Additionally, the alternate user mapping system allows for further flexibility on how to map tokens to iRODS users.

---
### OAuth Protected Resource #155
This allows access to the HTTP API endpoints by passing an access token as a Bearer token, as is specified in the OAuth specification: https://www.rfc-editor.org/rfc/rfc6749.html#section-7.1

Taking directly from the example provided in the above link, you would just need to send the access token as bearer:
```
GET /resource/1 HTTP/1.1
     Host: example.com
     Authorization: Bearer mF_9.B5f-4.1JqM
```

By sending a valid Bearer token, you should have access to the iRODS HTTP API endpoints. It should be noted that access will only be permitted in the following conditions:

1. The access token provided is valid.
2. The token can be matched to an iRODS user.

### Confidential Client Support #235
If you define `client_secret` in your configuration, this will run the HTTP API as a confidential client. This should allow for access to protected endpoints on the provider, such as the introspection endpoint, among other benefits.

When running as a confidential client, all requests to the OpenID Provider, aside from well-known endpoint, will send a Bearer token in the Authorization header with the provided `client_id` and `client_secret`, as described in the OAuth 2.0 specification: https://www.rfc-editor.org/rfc/rfc6749.html#section-2.3.1

### User Mapping #108
Without a mapping mechanism of access tokens to iRODS users, the HTTP API would not be able to function. By specifying the iRODS user, as well as the desired mapping attributes to search for. An example of a mapping can be found below:

```json
"openid_connect": {
  "user_attribute_mapping": {
    "rodsBob": {
      "email": "bob@bobtopia.example",
      "sub": "a.very.real.sub",
      "phone_number": "56709"
    },
    "rodsAlice": {
      "email": "al-1s@wonderland.example",
      "sub": "a.different.sub"
    }
  }
}
```

It should be noted that each iRODS user does not require the same number of attributes. Each of the given attributes must match completely in order for the token to be mapped to a user.

When running as a Protected Resource, these mappings will be matched with the response given from the OpenID Provider's introspection endpoint. An example of what may be in an introspection endpoint's response is as follows:

```json
 {
      "active": true,
      "client_id": "l238j323ds-23ij4",
      "username": "jdoe",
      "scope": "read write dolphin",
      "sub": "Z5O3upPC88QrAjx00dis",
      "aud": "https://protected.example.net/resource",
      "iss": "https://server.example.com/",
      "exp": 1419356238,
      "iat": 1419350238,
      "extension_field": "twenty-seven"
 }
```

The previous example was taken from the OAuth 2.0 Token Introspection specification, which can be found at the following link: https://www.rfc-editor.org/rfc/rfc7662.html#section-2.2

Otherwise, they will match against the OpenID Provider's ID Token, given in response to a token endpoint request. An example of what may be in an ID Token is as follows:

```json
{
   "iss": "https://server.example.com",
   "sub": "24400320",
   "aud": "s6BhdRkqt3",
   "nonce": "n-0S6_WzA2Mj",
   "exp": 1311281970,
   "iat": 1311280970,
   "auth_time": 1311280969,
   "acr": "urn:mace:incommon:iap:silver"
}
```

The previous example was taken from the OpenID Connect Core 1.0 specification, which can be found in the following link: https://openid.net/specs/openid-connect-core-1_0.html#IDToken

### DIFF
--- a/API.md
+++ b/API.md
@@ -30,12 +30,16 @@ Reauthentication can be performed at anytime and will result in a brand new toke
 
 ### Scheme: OpenID Connect (OIDC)
 
-For authenticating with OpenID Connect, there are two methods:
+For authenticating with OpenID Connect, there are three methods, two of which are run as clients:
 
-- Resource Owner Password Credentials Grant
-- Authorization Code Grant
+- Client-Based
+  - [Resource Owner Password Credentials Grant](#resource-owner-password-credentials-grant)
+  - [Authorization Code Grant](#authorization-code-grant)
+- [Protected Resource](#irods-as-an-oauth-protected-resource)
 
 #### Resource Owner Password Credentials Grant
+
+To run this client-based scheme, you must set `mode` in the `openid_connect` stanza to `client`.
 
 The core advantage of this grant is the flexibility in how it may be applied.
 
@@ -52,6 +56,8 @@ curl -X POST -H "Authorization: iRODS $username_and_password" \
 A string representing a bearer token that can be used to execute operations as the authenticated user.
 
 #### Authorization Code Grant
+
+To run this client-based scheme, you must set `mode` in the `openid_connect` stanza to `client`.
 
 Using this grant requires a bit more work to extract the token.
 Authentication is done in the browser.
@@ -83,6 +89,20 @@ This will open a browser window that will allow you to authenticate in.
 
 ##### Response
 The bearer token should be returned and viewable in the browser window after authenticating.
+
+#### iRODS as an OAuth Protected Resource
+To run as an OAuth protected resource, set `mode` in the `openid_connect` stanza to `protected_resource`.
+
+While running in this mode, the HTTP API does not provide any grants to which you can authenticate.
+Instead, you must implement an OAuth client yourself, authenticate, and provide an access token as a
+Bearer token when querying the HTTP API endpoints.
+
+In order for the access token to be accepted, two conditions must be met:
+1. The OpenID Provider recognizes the token as valid and active.
+2. The token must be able to be mapped to a user, using either `irods_user_claim` or `user_attribute_mapping`. 
+
+If these conditions are met, then the access token will be accepted, and the action will be carried out as the mapped
+user.
 
 ## Collection Operations
 
--- a/README.md
+++ b/README.md
@@ -263,6 +263,16 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g. 
                 // The client id given to the application by OIDC provider.
                 "client_id": "irods_http_api",
 
+                // The client secret used for accessing the introspection endpoint.
+                // Optional unless running as a protected resource.
+                "client_secret": "xxxxxxxxxxxxxxx",
+
+                // The OIDC mode the HTTP API will run as.
+                // The following values are supported:
+                // - client:              Run the HTTP API as an OIDC client
+                // - protected_resource:  Run as an OAuth Protected Resource
+                "mode": "client",
+
                 // URI pointing to the irods HTTP API auth endpoint.
                 "redirect_uri": "https://<domain>/irods-http-api/0.2.0/authenticate",
 
@@ -271,8 +281,19 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g. 
                 "state_timeout_in_seconds": 600,
 
                 // The name of the OIDC claim which provides the mapping of an
-                // OIDC user to an iRODS user account
+                // OIDC user to an iRODS user account.
+                // "irods_user_claim" and "user_attribute_mapping" cannot be used together.
                 "irods_user_claim": "irods_username",
+
+                // The mapping of a user to the provided values. All values must
+                // be matched to map an OIDC user to an iRODS user account.
+                // "irods_user_claim" and "user_attribute_mapping" cannot be used together.
+                "user_attribute_mapping": {
+                    "irods_username": {
+                        "sub": "123-abc-456-xyz",
+                        "email": "rods_user@example.org"
+                    }
+                },
 
                 // The path to the TLS certificates directory.
                 // Used for HTTPS connections.
--- a/core/include/irods/private/http_api/common.hpp
+++ b/core/include/irods/private/http_api/common.hpp
@@ -16,6 +16,8 @@
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
 
+#include <nlohmann/json.hpp>
+
 #include <chrono>
 #include <memory>
 #include <optional>
@@ -44,6 +46,7 @@ namespace irods::http
 	using request_handler_map_type = std::unordered_map<std::string_view, request_handler_type>;
 
 	using query_arguments_type = std::unordered_map<std::string, std::string>;
+	using body_arguments = std::unordered_map<std::string, std::string>;
 
 	using handler_type = void (*)(session_pointer_type, request_type&, query_arguments_type&);
 	// clang-format on
@@ -154,6 +157,17 @@ namespace irods::http
 	auto parse_url(const std::string& _url) -> url;
 
 	auto parse_url(const request_type& _req) -> url;
+
+	auto url_encode_body(const body_arguments& _args) -> std::string;
+
+	auto safe_base64_encode(std::string_view _view) -> std::string;
+
+	auto create_host_field(boost::urls::url_view _url, std::string_view _port) -> std::string;
+
+	auto create_oidc_request(boost::urls::url_view _url)
+		-> boost::beast::http::request<boost::beast::http::string_body>;
+
+	auto map_json_to_user(const nlohmann::json& _json) -> std::optional<std::string>;
 
 	auto resolve_client_identity(const request_type& _req) -> client_identity_resolution_result;
 
--- a/core/src/common.cpp
+++ b/core/src/common.cpp
@@ -5,8 +5,10 @@
 #include "irods/private/http_api/log.hpp"
 #include "irods/private/http_api/process_stash.hpp"
 #include "irods/private/http_api/session.hpp"
+#include "irods/private/http_api/transport.hpp"
 #include "irods/private/http_api/version.hpp"
 
+#include <irods/base64.hpp>
 #include <irods/client_connection.hpp>
 #include <irods/irods_at_scope_exit.hpp>
 #include <irods/irods_exception.hpp>
@@ -18,12 +20,21 @@
 #include <irods/ticketAdmin.h>
 
 #include <boost/any.hpp>
+#include <boost/asio.hpp>
 #include <boost/algorithm/string.hpp>
+#include <boost/beast.hpp>
 
 #include <curl/curl.h>
 #include <nlohmann/json.hpp>
 #include <spdlog/spdlog.h>
+
 #include <string>
+#include <string_view>
+
+// clang-format off
+namespace beast = boost::beast; // from <boost/beast.hpp>
+namespace net   = boost::asio;  // from <boost/asio.hpp>
+// clang-format on
 
 namespace irods::http
 {
@@ -207,6 +218,159 @@ namespace irods::http
 		return parse_url(fmt::format("http://ignored{}", _req.target()));
 	} // parse_url
 
+	auto url_encode_body(const body_arguments& _args) -> std::string
+	{
+		auto encode_pair{[](const body_arguments::value_type& i) {
+			return fmt::format("{}={}", encode(i.first), encode(i.second));
+		}};
+
+		return std::transform_reduce(
+			std::next(std::cbegin(_args)),
+			std::cend(_args),
+			encode_pair(*std::cbegin(_args)),
+			[](const auto& a, const auto& b) { return fmt::format("{}&{}", a, b); },
+			encode_pair);
+	}
+
+	auto safe_base64_encode(std::string_view _view) -> std::string
+	{
+		namespace logging = irods::http::log;
+
+		constexpr auto char_per_byte_set{4};
+		constexpr auto byte_set_size{3};
+
+		const auto max_size{char_per_byte_set * ((_view.size() + 2) / byte_set_size)};
+		auto max_size_plus_null_term{max_size + 1};
+
+		std::string encoded_data;
+		encoded_data.resize(max_size);
+
+		auto res{irods::base64_encode(
+			reinterpret_cast<const unsigned char*>(_view.data()),
+			_view.size(),
+			reinterpret_cast<unsigned char*>(encoded_data.data()),
+			&max_size_plus_null_term)};
+		if (res) {
+			logging::error("{}: Failed to encode the string [{}], output may be unusable.", __func__, _view);
+		}
+		return encoded_data;
+	}
+
+	auto create_host_field(boost::urls::url_view _url, std::string_view _port) -> std::string
+	{
+		if ((_port == "443" && _url.scheme_id() == boost::urls::scheme::https) ||
+		    (_port == "80" && _url.scheme_id() == boost::urls::scheme::http))
+		{
+			return _url.host();
+		}
+		return fmt::format("{}:{}", _url.host(), _port);
+	}
+
+	auto create_oidc_request(boost::urls::url_view _url) -> beast::http::request<beast::http::string_body>
+	{
+		constexpr auto http_version_number{11};
+		beast::http::request<beast::http::string_body> req{beast::http::verb::post, _url.path(), http_version_number};
+
+		const auto port{get_port_from_url(_url)};
+
+		req.set(beast::http::field::host, create_host_field(_url, *port));
+		req.set(beast::http::field::user_agent, irods::http::version::server_name);
+		req.set(beast::http::field::content_type, "application/x-www-form-urlencoded");
+		req.set(beast::http::field::accept, "application/json");
+
+		if (const auto secret_key{irods::http::globals::oidc_configuration().find("client_secret")};
+		    secret_key != std::end(irods::http::globals::oidc_configuration()))
+		{
+			const auto format_bearer_token{[](std::string_view _client_id, std::string_view _client_secret) {
+				auto encode_me{fmt::format("{}:{}", encode(_client_id), encode(_client_secret))};
+				return safe_base64_encode(encode_me);
+			}};
+
+			const auto& client_id{
+				irods::http::globals::oidc_configuration().at("client_id").get_ref<const std::string&>()};
+			const auto& client_secret{secret_key->get_ref<const std::string&>()};
+			const auto auth_string{fmt::format("Basic {}", format_bearer_token(client_id, client_secret))};
+
+			req.set(beast::http::field::authorization, auth_string);
+		}
+
+		return req;
+	}
+
+	auto hit_introspection_endpoint(std::string _encoded_body) -> nlohmann::json
+	{
+		namespace logging = irods::http::log;
+
+		const auto introspection_endpoint{irods::http::globals::oidc_endpoint_configuration()
+		                                      .at("introspection_endpoint")
+		                                      .get_ref<const std::string&>()};
+
+		const auto parsed_uri{boost::urls::parse_uri(introspection_endpoint)};
+
+		if (parsed_uri.has_error()) {
+			logging::error(
+				"{}: Error trying to parse introspection_endpoint [{}]. Please check configuration.",
+				__func__,
+				introspection_endpoint);
+			return {{"error", "bad endpoint"}};
+		}
+
+		const auto url{*parsed_uri};
+		const auto port{get_port_from_url(url)};
+
+		// Addr
+		net::io_context io_ctx;
+		auto tcp_stream{irods::http::transport_factory(url.scheme_id(), io_ctx)};
+		tcp_stream->connect(url.host(), *port);
+
+		// Build Request
+		auto req{create_oidc_request(url)};
+		req.body() = std::move(_encoded_body);
+		req.prepare_payload();
+
+		// Send request & receive response
+		auto res{tcp_stream->communicate(req)};
+
+		logging::debug("{}: Received the following response: [{}]", __func__, res.body());
+
+		// JSONize response
+		return nlohmann::json::parse(res.body());
+	}
+
+	auto map_json_to_user(const nlohmann::json& _json) -> std::optional<std::string>
+	{
+		const auto& oidc_config{irods::http::globals::oidc_configuration()};
+		const static auto user_claim{oidc_config.find("irods_user_claim")};
+		const static auto attribute_mapping{oidc_config.find("user_attribute_mapping")};
+
+		if (user_claim != std::end(oidc_config)) {
+			if (auto claim{_json.find(user_claim->get_ref<const std::string&>())}; claim != std::end(_json)) {
+				return {*claim};
+			}
+		}
+		else if (attribute_mapping != std::end(oidc_config)) {
+			const auto& mappings{*attribute_mapping};
+
+			for (auto& [key, value] : mappings.items()) {
+				const auto comparison_func{[&_json](const auto& _iter) -> bool {
+					const auto value_of_interest{_json.find(_iter.key())};
+					if (value_of_interest == std::end(_json)) {
+						return false;
+					}
+
+					return *value_of_interest == _iter.value();
+				}};
+
+				if (auto proxy_iter{value.items()};
+				    std::all_of(std::begin(proxy_iter), std::end(proxy_iter), comparison_func)) {
+					return key;
+				}
+			}
+		}
+
+		return std::nullopt;
+	}
+
 	auto resolve_client_identity(const request_type& _req) -> client_identity_resolution_result
 	{
 		namespace logging = irods::http::log;
@@ -237,6 +401,29 @@ namespace irods::http
 		// Verify the bearer token is known to the server. If not, return an error.
 		auto mapped_value{irods::http::process_stash::find(bearer_token)};
 		if (!mapped_value.has_value()) {
+			// If we're running as a protected resource, assume we have a OIDC token
+			if (irods::http::globals::oidc_configuration().at("mode").get_ref<const std::string&>() ==
+			    "protected_resource") {
+				body_arguments args{{"token", bearer_token}, {"token_type_hint", "access_token"}};
+
+				auto json_res{hit_introspection_endpoint(url_encode_body(args))};
+
+				// Validate access token
+				if (!json_res.at("active").get<bool>()) {
+					logging::warn("{}: Access token is invalid or expired.", __func__);
+					return {.response = fail(status_type::unauthorized)};
+				}
+
+				// Do mapping of user to irods user
+				auto user{map_json_to_user(json_res)};
+				if (user) {
+					return {.client_info = {.username = *std::move(user)}};
+				}
+
+				logging::warn("{}: Could not find a matching user.", __func__);
+				return {.response = fail(status_type::unauthorized)};
+			}
+
 			logging::error("{}: Could not find bearer token matching [{}].", __func__, bearer_token);
 			return {.response = fail(status_type::unauthorized)};
 		}
--- a/core/src/main.cpp
+++ b/core/src/main.cpp
@@ -239,29 +239,72 @@ constexpr auto default_jsonschema() -> std::string_view
                                     "minimum": 1
                                 }},
                                 "provider_url": {{
-                                    "type": "string"
+                                    "type": "string",
+                                    "format": "uri"
+                                }},
+                                "mode": {{
+                                    "enum": ["client", "protected_resource"]
                                 }},
                                 "client_id": {{
                                     "type": "string"
                                 }},
-                                "redirect_uri": {{
+                                "client_secret": {{
                                     "type": "string"
+                                }},
+                                "redirect_uri": {{
+                                    "type": "string",
+                                    "format": "uri"
                                 }},
                                 "irods_user_claim": {{
                                     "type": "string"
                                 }},
                                 "tls_certificates_directory": {{
                                     "type": "string"
+                                }},
+                                "user_attribute_mapping": {{
+                                    "type": "object",
+                                    "additionalProperties": {{
+                                        "type": "object"
+                                    }},
+                                    "minProperties": 1
                                 }}
                             }},
                             "required": [
                                 "timeout_in_seconds",
                                 "state_timeout_in_seconds",
                                 "provider_url",
+                                "mode",
                                 "client_id",
                                 "redirect_uri",
-                                "irods_user_claim",
                                 "tls_certificates_directory"
+                            ],
+                            "oneOf": [
+                                {{
+                                    "required": [
+                                        "irods_user_claim"
+                                    ]
+                                }},
+                                {{
+                                    "required": [
+                                        "user_attribute_mapping"
+                                    ]
+                                }}
+                            ],
+                            "anyOf": [
+                                {{
+                                    "not": {{
+                                        "properties": {{
+                                            "mode": {{
+                                                "const": "protected_resource"
+                                            }}
+                                        }}
+                                    }}
+                                }},
+                                {{
+                                    "required": [
+                                        "client_secret"
+                                    ]
+                                }}
                             ]
                         }}
                     }},
@@ -485,6 +528,8 @@ auto print_configuration_template() -> void
                 "state_timeout_in_seconds": 3600,
                 "provider_url": "<string>",
                 "client_id": "<string>",
+                "client_secret": "<string>",
+                "mode": "client",
                 "redirect_uri": "<string>",
                 "irods_user_claim": "<string>",
                 "tls_certificates_directory": "<string>"
@@ -779,9 +824,9 @@ auto load_oidc_configuration(const json& _config, json& _oi_config, json& _endpo
 		tcp_stream->connect(url.host(), *port);
 
 		// Build Request
-		constexpr auto version_number{11};
-		beast::http::request<beast::http::string_body> req{beast::http::verb::get, path, version_number};
-		req.set(beast::http::field::host, fmt::format("{}:{}", url.host(), *port));
+		constexpr auto http_version_number{11};
+		beast::http::request<beast::http::string_body> req{beast::http::verb::get, path, http_version_number};
+		req.set(beast::http::field::host, irods::http::create_host_field(url, *port));
 		req.set(beast::http::field::user_agent, irods::http::version::server_name);
 
 		// Sends and recieves response
--- a/endpoints/authentication/src/main.cpp
+++ b/endpoints/authentication/src/main.cpp
@@ -46,10 +46,19 @@ namespace beast = boost::beast; // from <boost/beast.hpp>
 namespace net   = boost::asio;  // from <boost/asio.hpp>
 // clang-format on
 
-using body_arguments = std::unordered_map<std::string, std::string>;
-
 namespace irods::http::handler
 {
+	auto remove_client_from_body_if_confidential_client(body_arguments _body) -> body_arguments
+	{
+		const static bool has_client_secret{irods::http::globals::oidc_configuration().contains("client_secret")};
+
+		if (has_client_secret) {
+			_body.erase("client_id");
+		}
+
+		return _body;
+	}
+
 	auto hit_token_endpoint(std::string _encoded_body) -> nlohmann::json
 	{
 		const auto token_endpoint{
@@ -68,22 +77,15 @@ namespace irods::http::handler
 
 		const auto url{*parsed_uri};
 		const auto port{irods::http::get_port_from_url(url)};
+		auto req{irods::http::create_oidc_request(url)};
 
-		// TCP thing
-		auto tcp_stream{irods::http::transport_factory(url.scheme_id(), io_ctx)};
-		tcp_stream->connect(url.host(), *port);
-
-		// Build Request
-		constexpr auto version_number{11};
-		beast::http::request<beast::http::string_body> req{beast::http::verb::post, url.path(), version_number};
-		req.set(beast::http::field::host, url.host());
-		req.set(beast::http::field::user_agent, irods::http::version::server_name);
-		req.set(beast::http::field::content_type,
-		        "application/x-www-form-urlencoded"); // Possibly set a diff way?
-
-		// Send
+		// Attach body to request
 		req.body() = std::move(_encoded_body);
 		req.prepare_payload();
+
+		// Connect to remote
+		auto tcp_stream{irods::http::transport_factory(url.scheme_id(), io_ctx)};
+		tcp_stream->connect(url.host(), *port);
 
 		// Send request and Read back response
 		auto res{tcp_stream->communicate(req)};
@@ -91,20 +93,6 @@ namespace irods::http::handler
 
 		// JSONize response
 		return nlohmann::json::parse(res.body());
-	}
-
-	auto encode_body(const body_arguments& _args) -> std::string
-	{
-		auto encode_pair{[](const body_arguments::value_type& i) {
-			return fmt::format("{}={}", irods::http::encode(i.first), irods::http::encode(i.second));
-		}};
-
-		return std::transform_reduce(
-			std::next(std::cbegin(_args)),
-			std::cend(_args),
-			encode_pair(*std::cbegin(_args)),
-			[](const auto& a, const auto& b) { return fmt::format("{}&{}", a, b); },
-			encode_pair);
 	}
 
 	auto is_error_response(const nlohmann::json& _response_to_check) -> bool
@@ -163,13 +151,24 @@ namespace irods::http::handler
 		return {std::move(username), std::move(password)};
 	}
 
+	auto is_oidc_running_as_client() -> bool
+	{
+		static const auto oidc_stanza_exists = irods::http::globals::configuration().contains(
+			nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"});
+
+		if (oidc_stanza_exists) {
+			static const auto is_client{
+				irods::http::globals::oidc_configuration().at("mode").get_ref<const std::string&>() == "client"};
+			return is_client;
+		}
+
+		return false;
+	}
+
 	IRODS_HTTP_API_ENDPOINT_ENTRY_FUNCTION_SIGNATURE(authentication)
 	{
 		if (_req.method() == boost::beast::http::verb::get) {
-			static const auto oidc_stanza_exists = irods::http::globals::configuration().contains(
-				nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"});
-
-			if (!oidc_stanza_exists) {
+			if (!is_oidc_running_as_client()) {
 				log::error("{}: HTTP GET method cannot be used for Basic authentication.", __func__);
 				return _sess_ptr->send(fail(status_type::method_not_allowed));
 			}
@@ -202,7 +201,7 @@ namespace irods::http::handler
 					const auto auth_endpoint{irods::http::globals::oidc_endpoint_configuration()
 					                             .at("authorization_endpoint")
 					                             .get_ref<const std::string&>()};
-					const auto encoded_url{fmt::format("{}?{}", auth_endpoint, encode_body(args))};
+					const auto encoded_url{fmt::format("{}?{}", auth_endpoint, irods::http::url_encode_body(args))};
 
 					log::debug("{}: Proper redirect to [{}]", fn, encoded_url);
 
@@ -314,7 +313,8 @@ namespace irods::http::handler
 					     irods::http::globals::oidc_configuration().at("redirect_uri").get_ref<const std::string&>()}};
 
 					// Encode the string, hit endpoint, get res
-					nlohmann::json oidc_response{hit_token_endpoint(encode_body(args))};
+					nlohmann::json oidc_response{hit_token_endpoint(
+						irods::http::url_encode_body(remove_client_from_body_if_confidential_client(args)))};
 
 					// Determine if we have an "error" json...
 					if (is_error_response(oidc_response)) {
@@ -329,11 +329,9 @@ namespace irods::http::handler
 					// TODO: Handle case where we throw!!!
 					auto decoded_token{jwt::decode<jwt::traits::nlohmann_json>(jwt_token).get_payload_json()};
 
-					// Verify 'irods_username' exists
-					const auto& irods_claim_name{irods::http::globals::oidc_configuration()
-					                                 .at("irods_user_claim")
-					                                 .get_ref<const std::string&>()};
-					if (!decoded_token.contains(irods_claim_name)) {
+					auto irods_username{irods::http::map_json_to_user(decoded_token)};
+
+					if (!irods_username) {
 						const auto user{
 							decoded_token.contains("preferred_username")
 								? decoded_token.at("preferred_username").get<const std::string>()
@@ -342,9 +340,6 @@ namespace irods::http::handler
 						log::error("{}: No irods user associated with authenticated user [{}].", fn, user);
 						return _sess_ptr->send(fail(status_type::bad_request));
 					}
-
-					// Get irods username
-					const std::string& irods_name{decoded_token.at(irods_claim_name).get_ref<const std::string&>()};
 
 					// Issue token?
 					static const auto seconds =
@@ -355,7 +350,7 @@ namespace irods::http::handler
 
 					auto bearer_token = irods::http::process_stash::insert(authenticated_client_info{
 						.auth_scheme = authorization_scheme::openid_connect,
-						.username = std::move(irods_name),
+						.username = *std::move(irods_username),
 						.expires_at = std::chrono::steady_clock::now() + std::chrono::seconds{seconds}});
 
 					response_type res_rep{status_type::ok, _req.version()};
@@ -523,7 +518,9 @@ namespace irods::http::handler
 					return _sess_ptr->send(std::move(res));
 				}
 				// OAuth 2.0 Resource Owner Password Credentials Grant
-				else if (const auto alt_method{iter->value().find("iRODS ")}; alt_method != std::string_view::npos) {
+				else if (const auto alt_method{iter->value().find("iRODS ")};
+				         is_oidc_running_as_client() && alt_method != std::string_view::npos)
+				{
 					// Decode username and password here!!!!!
 					constexpr auto basic_auth_scheme_prefix_size = 6;
 					const auto [username, password]{
@@ -543,7 +540,8 @@ namespace irods::http::handler
 						{"password", password}};
 
 					// Query endpoint
-					nlohmann::json oidc_response{hit_token_endpoint(encode_body(args))};
+					nlohmann::json oidc_response{hit_token_endpoint(
+						irods::http::url_encode_body(remove_client_from_body_if_confidential_client(args)))};
 
 					// Determine if we have an "error" json...
 					if (is_error_response(oidc_response)) {
@@ -555,12 +553,9 @@ namespace irods::http::handler
 
 					// Feed to JWT parser
 					auto decoded_token{jwt::decode<jwt::traits::nlohmann_json>(jwt_token).get_payload_json()};
+					auto irods_username{irods::http::map_json_to_user(decoded_token)};
 
-					// Verify 'irods_username' exists
-					const auto& irods_claim_name{irods::http::globals::oidc_configuration()
-					                                 .at("irods_user_claim")
-					                                 .get_ref<const std::string&>()};
-					if (!decoded_token.contains(irods_claim_name)) {
+					if (!irods_username) {
 						const auto user{
 							decoded_token.contains("preferred_username")
 								? decoded_token.at("preferred_username").get<const std::string>()
@@ -570,9 +565,6 @@ namespace irods::http::handler
 						return _sess_ptr->send(fail(status_type::bad_request));
 					}
 
-					// Get irods username
-					const std::string& irods_name{decoded_token.at(irods_claim_name).get_ref<const std::string&>()};
-
 					// Issue token?
 					static const auto seconds =
 						irods::http::globals::configuration()
@@ -581,7 +573,7 @@ namespace irods::http::handler
 							.get<int>();
 					auto bearer_token = irods::http::process_stash::insert(authenticated_client_info{
 						.auth_scheme = authorization_scheme::openid_connect,
-						.username = std::move(irods_name),
+						.username = *std::move(irods_username),
 						.expires_at = std::chrono::steady_clock::now() + std::chrono::seconds{seconds}});
 
 					response_type res_rep{status_type::ok, _req.version()};
--- a/test/config.py
+++ b/test/config.py
@@ -1,4 +1,5 @@
 import logging
+from jsonschema import validate
 
 test_config = {
     'log_level': logging.INFO,
@@ -6,6 +7,10 @@ test_config = {
     'host': 'localhost',
     'port': 9000,
     'url_base': '/irods-http-api/0.2.0',
+
+    'openid_connect': {
+        'mode': 'client'
+    },
 
     'rodsadmin': {
         'username': 'rods',
@@ -20,5 +25,73 @@ test_config = {
     'irods_zone': 'tempZone',
     'irods_server_hostname': 'localhost',
 
-    "run_genquery2_tests": False
+    'run_genquery2_tests': False
 }
+
+schema = {
+    '$schema': 'http://json-schema.org/draft-07/schema#',
+    '$id': 'https://schemas.irods.org/irods-http-api/test/0.2.0/test-schema.json',
+    'type': 'object',
+    'properties': {
+        'host': {
+            'type': 'string'
+        },
+        'port': {
+            'type': 'number'
+        },
+        'url_base': {
+            'type': 'string'
+        },
+        'openid_connect': {
+            'type': 'object',
+            'properties': {
+                'mode': {
+                    'enum': [ 'client', 'protected_resource' ]
+                }
+            },
+            'required': [ 'mode' ]
+        },
+        'rodsadmin': {
+            '$ref': '#/definitions/login'
+        },
+        'rodsuser': {
+            '$ref': '#/definitions/login'
+        },
+        'irods_zone': {
+            'type': 'string'
+        },
+        'irods_server_hostname': {
+            'type': 'string'
+        },
+        'run_genquery2_tests': {
+            'type': 'boolean'
+        }
+    },
+    'required': [
+        'host',
+        'port',
+        'url_base',
+        'openid_connect',
+        'rodsadmin',
+        'rodsuser',
+        'irods_zone',
+        'irods_server_hostname',
+        'run_genquery2_tests'
+    ],
+    'definitions': {
+        'login': {
+            'type': 'object',
+            'properties': {
+                'username': {
+                    'type': 'string'
+                },
+                'password': {
+                    'type': 'string'
+                }
+            },
+            'required': [ 'username', 'password' ]
+        }
+    }
+}
+
+validate(instance=test_config, schema=schema)
--- a/test/test_auth_endpoint.py
+++ b/test/test_auth_endpoint.py
@@ -1,3 +1,4 @@
+from config import test_config
 import pytest
 import requests
 import base64
@@ -5,32 +6,41 @@ import re
 import logging
 import html
 
-# Pragmatic
-# Search for IRODS_HTTP_API_BASE_URL
-# and IRODS_CLIENT_VERSION
-# TOOD: Consider turning this into a fixture
-AUTH_URL = 'http://127.0.0.1:9000/irods-http-api/0.2.0/authenticate'
+skip_protected_resource = pytest.mark.skipif(
+    test_config['openid_connect']['mode'] == 'protected_resource', reason='Not supported in "protected_resource" mode.')
 
-@pytest.mark.parametrize("username, password, expected_result", [('bob', 'bob', requests.codes.ok),
-                                                                 ('', '', requests.codes.unauthorized),
+skip_client = pytest.mark.skipif(
+    test_config['openid_connect']['mode'] == 'client', reason='Not supported in "client" mode.')
+
+@pytest.fixture
+def irods_http_api_url_base():
+    return f'http://{test_config["host"]}:{test_config["port"]}{test_config["url_base"]}'
+
+@pytest.fixture
+def auth_url(irods_http_api_url_base):
+    return f'{irods_http_api_url_base}/authenticate'
+
+@pytest.mark.parametrize("username, password, expected_result", [pytest.param('bob', 'bob', requests.codes.ok, marks=skip_protected_resource),
+                                                                 pytest.param('', '', requests.codes.unauthorized, marks=skip_protected_resource),
                                                                  ('not', 'valid', requests.codes.bad_request),
-                                                                 ('a'*200, 'b'*200, requests.codes.unauthorized)])
-def test_oidc_resource_owner_password_credentials_login(username, password, expected_result):
+                                                                 pytest.param('a'*200, 'b'*200, requests.codes.unauthorized, marks=skip_protected_resource)])
+def test_oidc_resource_owner_password_credentials_login(username, password, expected_result, auth_url):
     to_be_encoded = f'{username}:{password}'
     encoded_user_pass = base64.b64encode(to_be_encoded.encode())
 
-    res = requests.post(AUTH_URL, headers={'Authorization': f'iRODS {encoded_user_pass.decode()}'})
+    res = requests.post(auth_url, headers={'Authorization': f'iRODS {encoded_user_pass.decode()}'})
 
     # Got a good code, assume we passed...
     assert res.status_code == expected_result
 
-def test_oidc_authorization_code_login():
+@skip_protected_resource
+def test_oidc_authorization_code_login(auth_url):
     # Session required for cookie
     s = requests.Session()
 
     # Go to login page
     # requests automatically redirects
-    res = s.get(AUTH_URL)
+    res = s.get(auth_url)
     assert res.status_code == requests.codes.ok
 
     data = res.text
@@ -51,13 +61,14 @@ def test_oidc_authorization_code_login():
     # Make sure we weren't redirected back the the auth page
     assert re.match('<!DOCTYPE html>', res.text) is None
 
-def test_oidc_authorization_code_non_irods_user():
+@skip_protected_resource
+def test_oidc_authorization_code_non_irods_user(auth_url):
     # Session required for cookie
     s = requests.Session()
 
     # Go to login page
     # requests automatically redirects
-    res = s.get(AUTH_URL)
+    res = s.get(auth_url)
     assert res.status_code == requests.codes.ok
 
     data = res.text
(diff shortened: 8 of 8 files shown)


core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+
+	auto res{irods::base64_encode(reinterpret_cast<const unsigned char*>(_view.data()),
+								  _view.size(),
+								  reinterpret_cast<unsigned char*>(encoded_data.data()),
+								  &max_size_plus_null_term)};
+	if (res) {
+	  log::debug("{}: Yikes!!!", __func__);
+	}
+	return encoded_data;
+  }
```
MartinFlores751, line 252: Formatting. Also, need a better name for this.

core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+			return fmt::format("{}={}", encode(i.first), encode(i.second));
+		}};
+
+		return std::transform_reduce(
+			std::next(std::cbegin(_args)),
+			std::cend(_args),
+			encode_pair(*std::cbegin(_args)),
+			[](const auto& a, const auto& b) { return fmt::format("{}&{}", a, b); },
+			encode_pair);
+	}
```
MartinFlores751, line 233: Consider changing the function name?
Feel like it could be a tad more general.

core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+
+	auto is_uuid(std::string_view _token) -> bool {
+		boost::uuids::string_generator gen{};
+		try {
+			auto possible_uuid{gen(_token.data())}; // This throws
+			return possible_uuid.version() != boost::uuids::uuid::version_type::version_unknown;
+		} catch (...) {
+			return false;
+		}
+	}
```
MartinFlores751, line 314: Not used anymore, can be removed.

core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+				++first;
+			}
+			return last;
+		}
+
+		template <class InputIt, class UnaryPredicate>
+		constexpr auto all_of(InputIt first, InputIt last, UnaryPredicate p) -> bool {
+			return irods::http::json::find_if_not(first, last, p) == last;
+		}
+	} // namespace json
```
MartinFlores751, line 332: Artifacts of me wanting `all_of` to work with json iterators.

Really, only line 320 is what needed to be changed. We don't want a receive a `*first` as our argument.

Could always not do the functional way too.

core/src/common.cpp
```diff
@@ -237,6 +361,44 @@ namespace irods::http
...
+				body_arguments args{
+					{"token", bearer_token},
+					{"token_type_hint", "access_token"}};
+
+				auto json_res{hit_intro_endpoint(url_encode_body(args))};
+				log::debug("{}: Got the following resp back: [{}]", __func__, json_res);
+
+				// Do fun bearer token validation
+				if (!json_res.at("active").get<bool>()) {
+					log::error("{}: This BOZO thinks they can just waltz on in...", __func__);
```
MartinFlores751, line 375: Don't forget to change this helpful message to something a bit more useful…

endpoints/authentication/src/main.cpp
```diff
@@ -46,10 +46,54 @@ namespace beast = boost::beast; // from <boost/beast.hpp>
...
+
+	auto remove_client_from_body_if_confidential_client(body_arguments _body) -> body_arguments {
+		const static bool has_client_secret{irods::http::globals::oidc_configuration().contains("client_secret")};
+
+		if (has_client_secret) {
+			_body.erase("client_id");
+		}
+
+		return _body;
+	}
```
MartinFlores751, line 67: Still not sure if this is hacky or not.

endpoints/authentication/src/main.cpp
```diff
@@ -46,10 +46,54 @@ namespace beast = boost::beast; // from <boost/beast.hpp>
...
+
+	auto create_oidc_request(boost::urls::url_view _url) -> beast::http::request<beast::http::string_body> {
+		constexpr auto version_number{11};
+		beast::http::request<beast::http::string_body> req{beast::http::verb::post, _url.path(), version_number};
+
+		const auto port{irods::http::get_port_from_url(_url)};
+
+		req.set(beast::http::field::host, create_host_field(_url, *port)); // TODO: Add port if not matching default scheme
```
MartinFlores751, line 75: Remove TODO.

endpoints/authentication/src/main.cpp
```diff
@@ -163,13 +186,23 @@ namespace irods::http::handler
...
+	}
+
 	IRODS_HTTP_API_ENDPOINT_ENTRY_FUNCTION_SIGNATURE(authentication)
 	{
 		if (_req.method() == boost::beast::http::verb::get) {
-			static const auto oidc_stanza_exists = irods::http::globals::configuration().contains(
-				nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"});
-
-			if (!oidc_stanza_exists) {
+			if (!is_oidc_running_as_client()) {
```
MartinFlores751, line 205: I think this doesn't do what we want it to...

endpoints/authentication/src/main.cpp
```diff
@@ -523,7 +556,7 @@ namespace irods::http::handler
 					return _sess_ptr->send(std::move(res));
 				}
 				// OAuth 2.0 Resource Owner Password Credentials Grant
-				else if (const auto alt_method{iter->value().find("iRODS ")}; alt_method != std::string_view::npos) {
+				else if (const auto alt_method{iter->value().find("iRODS ")}; is_oidc_running_as_client() && alt_method != std::string_view::npos) {
```
MartinFlores751, line 559: I think this doesn't do what we want it to...

core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+		beast::http::request<beast::http::string_body> req{beast::http::verb::post, url.path(), version_number};
+		req.set(beast::http::field::host, url.host());
+		req.set(beast::http::field::user_agent, irods::http::version::server_name);
+		req.set(beast::http::field::content_type,
+		        "application/x-www-form-urlencoded"); // Possibly set a diff way?
+		req.set(beast::http::field::authorization, auth_string);
+		req.set(beast::http::field::accept, "application/json");
+
+		req.body() = std::move(_encoded_body);
+		req.prepare_payload();
```
MartinFlores751, line 295: Can we use `create_oidc_request(...)` here?

core/src/common.cpp
```diff
@@ -237,6 +361,44 @@ namespace irods::http
...
+						return false;
+					}};
+
+					if (irods::http::json::all_of(std::cbegin(inner_val), std::cend(inner_val), comparison_func)) {
+						return {.client_info = {.username = element.key()}};
+					}
+				}
+
+				log::error("{}: Could not find a matching user.", __func__);
+				return {.response = fail(status_type::unauthorized)};
```
MartinFlores751, line 399: If we're going to log & leave here, make it unique/useful, otherwise, just fall through to the next error.

MartinFlores751 (commented):
Noting what I do recall or have written down.

core/src/common.cpp
korydraughn: What is the problem you're trying to solve with this?

MartinFlores751: When you dereference a json iterator, you don't get the 'key' value, which is needed for mapping.

By not dereferencing the iterator in `find_if_not` on line 320, we get the information we need (a key-value pair).

Link to portion of code where I use the funtions defined:
https://github.com/irods/irods_client_http_api/blob/66e8f61475d0eed61bca5598a89475a91f1cd7db/core/src/common.cpp#L379-L396

core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+		  
+	std::string encoded_data{};
+	encoded_data.resize(max_size);
+
+	auto res{irods::base64_encode(reinterpret_cast<const unsigned char*>(_view.data()),
+								  _view.size(),
+								  reinterpret_cast<unsigned char*>(encoded_data.data()),
+								  &max_size_plus_null_term)};
+	if (res) {
+	  log::debug("{}: Yikes!!!", __func__);
```
trel, line 249: update debug message with more/better description

core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+	}
+	return encoded_data;
+  }
+  
+    auto hit_intro_endpoint(std::string _encoded_body) -> nlohmann::json
```
trel, line 254: let's use all the letters

intro -> introspection

core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+		  auto encode_me{fmt::format("{}:{}", encode(_client_id), encode(_client_secret))};
+		  return safe_base64_encode(encode_me);
+		}};
+
+		auto client_id{irods::http::globals::oidc_configuration().at("client_id").get_ref<const std::string&>()};
+		auto client_secret{irods::http::globals::oidc_configuration().at("client_secret").get_ref<const std::string&>()};
+		auto auth_string{fmt::format("Basic {}", format_bearer_token(client_id, client_secret))};
+	
+		// Build Request
+		constexpr auto version_number{11};
```
trel, line 285: what is 11?

core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+		req.set(beast::http::field::user_agent, irods::http::version::server_name);
+		req.set(beast::http::field::content_type,
+		        "application/x-www-form-urlencoded"); // Possibly set a diff way?
+		req.set(beast::http::field::authorization, auth_string);
+		req.set(beast::http::field::accept, "application/json");
+
+		req.body() = std::move(_encoded_body);
+		req.prepare_payload();
+
+		// Send request & recieve response
```
trel, line 297: recieve -> receive

core/src/common.cpp
```diff
@@ -207,6 +218,119 @@ namespace irods::http
...
+		auto res{tcp_stream->communicate(req)};
+		
+		log::debug("{}: Got the following resp back: [{}]", __func__, res.body());
```
trel, line 300: consider

"Received the following response:"

core/src/common.cpp
```diff
@@ -237,6 +361,44 @@ namespace irods::http
...
 		auto mapped_value{irods::http::process_stash::find(bearer_token)};
 		if (!mapped_value.has_value()) {
+			// If we're running as a protected resource, assume we have a OIDC token
+			if (irods::http::globals::oidc_configuration().at("run_as_protected_resource").get<bool>()) {
+				body_arguments args{
+					{"token", bearer_token},
+					{"token_type_hint", "access_token"}};
+
+				auto json_res{hit_intro_endpoint(url_encode_body(args))};
+				log::debug("{}: Got the following resp back: [{}]", __func__, json_res);
```
trel, line 371: consider

"Received the following response:"

core/src/common.cpp
```diff
@@ -237,6 +361,44 @@ namespace irods::http
...
+				// Do fun bearer token validation
+				if (!json_res.at("active").get<bool>()) {
+					log::error("{}: This BOZO thinks they can just waltz on in...", __func__);
+					return {.response = fail(status_type::unauthorized)};
+				}
+
+				// Do mapping of user to irods user
+				const auto& mappings{irods::http::globals::oidc_configuration().at("user_attribute_mapping")};
+
+				// O(n x m)?
```
trel, line 382: can we create a map and then do a simple lookup?  O(1)?

core/src/main.cpp
```diff
@@ -235,24 +235,41 @@ constexpr auto default_jsonschema() -> std::string_view
                                     "minimum": 1
                                 }},
                                 "provider_url": {{
-                                    "type": "string"
+                                    "type": "string",
+                                    "format": "uri"
+                                }},
+                                "run_as_protected_resource": {{
```
trel, line 241: discush... is this in addition to running as an oidc client?  or instead of?  do we want to create a 'mode' instead of a boolean for this?

core/src/main.cpp
```diff
@@ -235,24 +235,41 @@ constexpr auto default_jsonschema() -> std::string_view
...
+                                    "format": "uri"
+                                }},
                                 "irods_user_claim": {{
                                     "type": "string"
                                 }},
                                 "tls_certificates_directory": {{
                                     "type": "string"
+                                }},
+                                "user_attribute_mapping": {{
+                                    "description": "Bleh blah",
```
trel, line 261: not sure what this is or what it should be.

endpoints/authentication/src/main.cpp
trel: or substitute xxxxxxxxx.    depends on the use case for this function.     we do the same when logging configuration values in other projects / libraries.

endpoints/authentication/src/main.cpp
```diff
@@ -46,10 +46,54 @@ namespace beast = boost::beast; // from <boost/beast.hpp>
...
+
+		if (has_client_secret) {
+			_body.erase("client_id");
+		}
+
+		return _body;
+	}
+
+	auto create_oidc_request(boost::urls::url_view _url) -> beast::http::request<beast::http::string_body> {
+		constexpr auto version_number{11};
```
trel, line 70: magic number?    or perhaps leave a comment.

endpoints/authentication/src/main.cpp
```diff
@@ -68,23 +112,16 @@ namespace irods::http::handler
 
 		const auto url{*parsed_uri};
 		const auto port{irods::http::get_port_from_url(url)};
+		auto req{create_oidc_request(url)};
+
+		// Attach body to request
+		req.body() = std::move(_encoded_body);
+		req.prepare_payload();
 
 		// TCP thing
```
trel, line 121: tweak please

endpoints/authentication/src/main.cpp
```diff
@@ -163,13 +186,23 @@ namespace irods::http::handler
...
 	}
 
+	auto is_oidc_running_as_client() -> bool
+	{
+		static const auto oidc_stanza_exists = irods::http::globals::configuration().contains(
+			nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"});
+
+		if (oidc_stanza_exists) {
+			static const auto is_client{irods::http::globals::oidc_configuration().at("run_as_protected_resource").get<bool>()};
+			return is_client;
```
trel, line 196: okay, so this *is* an either/or situation...   so an alternative universe could have a 'mode' set in the configuration with an enum of ... ['client', 'protected_resource'].

trel (commented):
excited to get a demo

good stuff

core/src/common.cpp
MartinFlores751: Don't think we can. That would require the JSON in the mapping and the introspection response to exactly match, right?

The introspection response can return some 'time' data, so it won't be consistent.

endpoints/authentication/src/main.cpp
MartinFlores751: Pointing over yonder: https://github.com/irods/irods_client_http_api/pull/252#discussion_r1498533002

core/src/common.cpp
trel: Ah.   Yes, please leave comment - that's pretty opaque.

core/src/common.cpp
korydraughn: Correct me if I'm wrong, but the reason for the O(n * m) is because the mapping supports identifying a user through a combination of attributes/properties.

If the mapping was strictly something like the following, where `<oidc_atrribute_value>` is the value mapped to `email_address`, then we could get O(1). This is not as flexible as the what @MartinFlores751 has presented.
```js
{
  "<oidc_atrribute_value>": "alice",
  "<oidc_atrribute_value>": "rods"
  // ... more mappings ...
}
```

However, the style of mapping could also be an option. Anything's possible, it's just depends on the pros/cons.

core/src/main.cpp
korydraughn: Noting here that we discussed this and landed on this should isn't a high priority item.

Let's capture it in an issue and investigate whether it seems helpful to users. As of right now, it doesn't offer much benefit because the it gets lost in the validation error output produced by jsonschema.

core/src/main.cpp
korydraughn: An enum definitely gives us more flexibility.

core/src/common.cpp
trel: Makes sense.   Carry on...

endpoints/authentication/src/main.cpp
trel: or even make the oidc stanza required... but with a 'mode' that could be one of ['none', 'client', 'protected_resource'].

could provide flexibility for future good ideas.

endpoints/authentication/src/main.cpp
```diff
@@ -388,7 +347,7 @@ namespace irods::http::handler
 
 					auto bearer_token = irods::http::process_stash::insert(authenticated_client_info{
 						.auth_scheme = authorization_scheme::openid_connect,
-						.username = std::move(irods_name),
+						.username = *irods_username,
```
korydraughn, line 350: Perhaps a std::move can be done here?

endpoints/authentication/src/main.cpp
```diff
@@ -614,7 +567,7 @@ namespace irods::http::handler
 							.get<int>();
 					auto bearer_token = irods::http::process_stash::insert(authenticated_client_info{
 						.auth_scheme = authorization_scheme::openid_connect,
-						.username = std::move(irods_name),
+						.username = *irods_username,
```
korydraughn, line 570: std::move?

korydraughn (commented):
This is coming together nicely.

Found a few things.

core/include/irods/private/http_api/common.hpp
```diff
@@ -155,6 +158,18 @@ namespace irods::http
...
 
+	auto url_encode_body(const body_arguments& _args) -> std::string;
+
+	auto safe_base64_encode(std::string_view _view) -> std::string;
+
+	auto create_host_field(boost::urls::url_view _url, std::string_view _port) -> std::string;
+
+	auto create_oidc_request(boost::urls::url_view _url) -> boost::beast::http::request<boost::beast::http::string_body>;
+
+	auto is_uuid(std::string_view _token) -> bool;
```
korydraughn, line 169: We keeping or removing this?

core/src/common.cpp
```diff
@@ -207,6 +218,178 @@ namespace irods::http
...
+			encode_pair);
+	}
+
+  auto safe_base64_encode(std::string_view _view) -> std::string {
+	constexpr auto char_per_byte_set{4};
+	constexpr auto byte_set_size{3};
+	auto max_size{char_per_byte_set * ((_view.size() + 2) / byte_set_size)};
+	auto max_size_plus_null_term{max_size + 1};
+		  
+	std::string encoded_data{};
```
korydraughn, line 241: The curly braces are unnecessary. Please remove.

core/src/common.cpp
```diff
@@ -207,6 +218,178 @@ namespace irods::http
...
+
+		// Addr
+		net::io_context io_ctx;
+		auto tcp_stream{irods::http::transport_factory(url.scheme_id(), io_ctx)};
+		tcp_stream->connect(url.host(), *port);
+
+		const auto format_bearer_token{[](std::string_view _client_id, std::string_view _client_secret){
+		  auto encode_me{fmt::format("{}:{}", encode(_client_id), encode(_client_secret))};
+		  return safe_base64_encode(encode_me);
+		}};
```
korydraughn, line 314: I've seen this twice now. Consider moving into a function.

endpoints/authentication/src/main.cpp
korydraughn: Is this because of logging?

Alternatively, you could overwrite it with `[REDACTED]` or something.

endpoints/authentication/src/main.cpp
korydraughn: Please elaborate.

korydraughn (commented):
Just realized I reviewed a portion of the PR.

I think I covered it all now.

core/include/irods/private/http_api/common.hpp
MartinFlores751: We're tossing this one.

core/src/common.cpp
MartinFlores751: This lambda and some variables below it aren't used, since this is already taken care of in another function, `create_oidc_request(...)`.

core/src/common.cpp
```diff
@@ -250,7 +250,43 @@ namespace irods::http
 	}
 	return encoded_data;
   }
-  
+
+	auto create_host_field(boost::urls::url_view _url, std::string_view _port) -> std::string {
+		if ((_port.compare("443") == 0 && _url.scheme_id() == boost::urls::scheme::https) ||
+			(_port.compare("80") == 0 && _url.scheme_id() == boost::urls::scheme::http)) {
```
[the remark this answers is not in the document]
MartinFlores751: No particular reason from what I recall. Change it?

core/src/common.cpp
```diff
@@ -250,7 +250,43 @@ namespace irods::http
...
+			return _url.host();
+		}
+		return fmt::format("{}:{}", _url.host(), _port);
+	}
```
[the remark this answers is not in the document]
MartinFlores751: OpenID Providers, such as Keycloak, are sensitive to whether a port is present in the header. More broadly, HTTP applications can program around ports. Common practice, from what I can tell, is to send the port if it is non-standard for the scheme.

In the case of Keycloak, this makes all communications work without explicit overrides in the Keycloak configuration.

core/src/common.cpp
```diff
@@ -250,7 +250,43 @@ namespace irods::http
...
+		const auto port{get_port_from_url(_url)};
+
+		req.set(beast::http::field::host, create_host_field(_url, *port));
+		req.set(beast::http::field::user_agent, irods::http::version::server_name);
+		req.set(beast::http::field::content_type,
+		        "application/x-www-form-urlencoded");
+		req.set(beast::http::field::accept, "application/json");
+
+		if (const auto& g_config{irods::http::globals::oidc_configuration()}; g_config.contains("client_secret")) {
+			const auto format_bearer_token{[](std::string_view _client_id, std::string_view _client_secret){
```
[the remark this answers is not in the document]
MartinFlores751: No, but I do kinda like it since, in my opinion, it makes clear what it does 'format_bearer_token', and cleans up the initialization in the variable following the definition. Not opposed to removing, though.

MartinFlores751:
Response to: https://github.com/irods/irods_client_http_api/pull/252#discussion_r1501131318

Not for logging. OAuth spec allows you to auth either in the header (preferred) or the body. Some OpenID Providers may not appreciate finding a spare client_id in the body after finding it in the header.

MartinFlores751:
Response to: https://github.com/irods/irods_client_http_api/pull/252#discussion_r1501132405

`is_oidc_running_as_client()` is `oidc_stanza_exists` plus `run_as_protected_resource`. If we run as a protected resource, we don't run as a client, which is most of the auth endpoint.

core/src/common.cpp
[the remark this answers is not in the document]
korydraughn: It's fine for now.

core/src/common.cpp
MartinFlores751: Done.

core/src/main.cpp
MartinFlores751: Leave as is & create issue for this?

core/include/irods/private/http_api/common.hpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/main.cpp
korydraughn: Let's go ahead and handle this.

Post what you think the enum property should be called and what values it should accept.

core/src/common.cpp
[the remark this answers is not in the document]
korydraughn: Yes please. `==` is just easier to read.

core/src/common.cpp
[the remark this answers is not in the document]
korydraughn: Based on the impl, there doesn't seem to be a downside to always including the port. It's more explicit and doesn't leave room for assumptions.

Thoughts?

core/src/common.cpp
```diff
@@ -250,7 +250,43 @@ namespace irods::http
...
+				auto encode_me{fmt::format("{}:{}", encode(_client_id), encode(_client_secret))};
```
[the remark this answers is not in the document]
korydraughn: Does OIDC require the Bearer token (i.e. `<lhs>:<rhs>`) to be URL encoded?

I thought it only needed to be base64 encoded.

core/src/main.cpp
MartinFlores751: To clarify, do we expect an 'oidc_mode' and have it be 'protected_resource' or 'client' for example? Is this better than a boolean?

core/src/common.cpp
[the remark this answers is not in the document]
MartinFlores751: Maybe? I just want to lean on what's more common practice.

core/src/common.cpp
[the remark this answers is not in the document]
MartinFlores751: Yes, here's relevant part: https://www.rfc-editor.org/rfc/rfc6749.html#section-2.3.1

They want the user & pass to be url encoded, joined with a ':', and then base64 encoded.

core/src/main.cpp
korydraughn: Introducing a mode just gives us more flexibility if/when we need to change things. The boolean only gives us two values to work with.

core/src/common.cpp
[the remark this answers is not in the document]
korydraughn: And that's completely fair.

How is the port captured and passed to this function?

core/src/common.cpp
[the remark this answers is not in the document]
korydraughn: Got it. Thanks.

core/src/main.cpp
MartinFlores751: Gotcha. In that case, are we going to require the OIDC section, or have two ways to disable OIDC via 'none' and omission?

trel: after discussion...

- we'll use basic if it's defined
- we'll use openid_connect if it's defined

no conditional logic or 'mode' for now.

trel: and 'mode' for 'client' or 'protected_resource' within openid_connect.

endpoints/authentication/src/main.cpp
MartinFlores751: Think we're good on this one, unless we prefer the inverse, i.e. add client_id later compared to removing it?

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Made version number's purpose more clear.

core/src/common.cpp
MartinFlores751: Done.

core/src/main.cpp
MartinFlores751: Done, leaving open just in case we prefer a different name than 'mode'.

core/src/main.cpp
MartinFlores751: Captured in issue #262 .

endpoints/authentication/src/main.cpp
MartinFlores751: Renamed.

core/src/common.cpp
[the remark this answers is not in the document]
MartinFlores751: The port is taken from the url, whether that's from the config or the well-known response, which is parsed using boost url.

core/src/common.cpp
MartinFlores751: Cleaned up.

README.md
MartinFlores751, line 1: Don't forget to document all new options in the README!

endpoints/authentication/src/main.cpp
korydraughn: Just circling back to this. What are we trying to solve here? I need for context to understand the issue.

core/src/common.cpp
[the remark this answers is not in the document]
korydraughn: Ok. Letting it ride.

endpoints/authentication/src/main.cpp
MartinFlores751: This got separated since the replies in the reviews threw me off:

> Not for logging. OAuth spec allows you to auth either in the header (preferred) or the body. Some OpenID Providers may not appreciate finding a spare client_id in the body after finding it in the header.

I think this is the context you're looking for too?

korydraughn: I see. And so the approach you've chosen is to leave this as is _(i.e. remove it from the body)_?

MartinFlores751: Yep, not opposed to the other way, though.

endpoints/authentication/src/main.cpp
MartinFlores751: I thought there was a logic issue in `is_oidc_running_as_client(...)`.

endpoints/authentication/src/main.cpp
korydraughn: I don't have strong opinions on it.

Let's leave it as is for now.

endpoints/authentication/src/main.cpp
korydraughn: And so, what's the verdict now?

MartinFlores751: I'm going to double-check it, but I think it works. Besides, with the schema change, it needs to be updated anywho.

endpoints/authentication/src/main.cpp
MartinFlores751: Going to point this one here: https://github.com/irods/irods_client_http_api/pull/252#discussion_r1498535591

It'll be closed when the linked on is.

core/src/common.cpp
```diff
@@ -377,22 +436,9 @@ namespace irods::http
...
-						}
-						return false;
-					}};
-
-					if (irods::http::json::all_of(std::cbegin(inner_val), std::cend(inner_val), comparison_func)) {
-						return {.client_info = {.username = element.key()}};
-					}
+				auto user{map_json_to_user(json_res)};
+				if (user) {
+					return {.client_info = {.username = *user}};
```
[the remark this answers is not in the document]
MartinFlores751: Done, check to see if you like the method, though.

endpoints/authentication/src/main.cpp
MartinFlores751: Done, check to see if you like the method, though.

endpoints/authentication/src/main.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

endpoints/authentication/src/main.cpp
MartinFlores751: Done. No more 'TCP thing'.

core/src/common.cpp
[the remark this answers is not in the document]
MartinFlores751: Done.

endpoints/authentication/src/main.cpp
MartinFlores751: Works as expected.

trel: ah, very good.

core/src/common.cpp
```diff
@@ -250,7 +250,43 @@ namespace irods::http
...
+
+		const auto port{get_port_from_url(_url)};
+
+		req.set(beast::http::field::host, create_host_field(_url, *port));
+		req.set(beast::http::field::user_agent, irods::http::version::server_name);
+		req.set(beast::http::field::content_type,
+		        "application/x-www-form-urlencoded");
+		req.set(beast::http::field::accept, "application/json");
+
+		if (const auto& g_config{irods::http::globals::oidc_configuration()}; g_config.contains("client_secret")) {
```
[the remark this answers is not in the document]
MartinFlores751: Think it would be better to `static const bool` the `.contains(...)`, so it doesn't run multiple times?

Not sure if you can do the same with an itter.

core/src/common.cpp
```diff
@@ -331,6 +359,37 @@ namespace irods::http
 		}
 	} // namespace json
 
+	auto map_json_to_user(const nlohmann::json& _json) -> std::optional<std::string> {
+		const static bool use_user_claim{irods::http::globals::oidc_configuration().contains("irods_user_claim")};
+		const static bool use_attibute_mapping{irods::http::globals::oidc_configuration().contains("user_attribute_mapping")};
+
+		if (use_user_claim) {
+			if(auto claim{_json.find(irods::http::globals::oidc_configuration()["irods_user_claim"].get_ref<const std::string&>())}; claim != std::end(_json)) {
```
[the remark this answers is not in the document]
MartinFlores751: It's unchecked because we already checked with `use_user_claim`. This does not rely on schema validation.

core/src/common.cpp
```diff
@@ -331,6 +359,37 @@ namespace irods::http
...
+			const auto& mappings{irods::http::globals::oidc_configuration()["user_attribute_mapping"]};
+
+			for (auto& element : mappings.items()) {
+				const auto inner_val{element.value()};
+
+				const auto comparison_func{[&_json](const auto& _iter) -> bool {
+					if (_json.contains(_iter.key()) && _json[_iter.key()] == _iter.value()) {
+						return true;
+					}
+					return false;
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -331,6 +359,37 @@ namespace irods::http
...
+			}
+		}
+		else if (use_attibute_mapping) {
+			const auto& mappings{irods::http::globals::oidc_configuration()["user_attribute_mapping"]};
+
+			for (auto& element : mappings.items()) {
+				const auto inner_val{element.value()};
+
+				const auto comparison_func{[&_json](const auto& _iter) -> bool {
+					if (_json.contains(_iter.key()) && _json[_iter.key()] == _iter.value()) {
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -331,6 +359,37 @@ namespace irods::http
...
+						return true;
+					}
+					return false;
+				}};
+
+				if (irods::http::json::all_of(std::cbegin(inner_val), std::cend(inner_val), comparison_func)) {
+					return {element.key()};
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -331,6 +359,37 @@ namespace irods::http
...
+
+		if (use_user_claim) {
+			if(auto claim{_json.find(irods::http::globals::oidc_configuration()["irods_user_claim"].get_ref<const std::string&>())}; claim != std::end(_json)) {
+				return {*claim};
+			}
+		}
+		else if (use_attibute_mapping) {
+			const auto& mappings{irods::http::globals::oidc_configuration()["user_attribute_mapping"]};
+
+			for (auto& element : mappings.items()) {
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Turns out I was holding the json library wrong. Fixed.

core/src/common.cpp
MartinFlores751: Think this is better handled in a refactor/pass in the OIDC related items.

core/src/common.cpp
MartinFlores751: Think the log is different enough from the fall through/next error.

core/src/common.cpp
```diff
@@ -331,6 +359,37 @@ namespace irods::http
 		}
 	} // namespace json
 
+	auto map_json_to_user(const nlohmann::json& _json) -> std::optional<std::string> {
+		const static bool use_user_claim{irods::http::globals::oidc_configuration().contains("irods_user_claim")};
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/main.cpp
```diff
@@ -465,7 +508,9 @@ auto print_configuration_template() -> void
             "openid_connect": {{
                 "timeout_in_seconds": 3600,
                 "provider_url": "<string>",
+                "run_as_protected_resource": "<string>",
```
korydraughn, line 511: Is this still needed?

endpoints/authentication/src/main.cpp
```diff
@@ -355,7 +347,7 @@ namespace irods::http::handler
 
 					auto bearer_token = irods::http::process_stash::insert(authenticated_client_info{
 						.auth_scheme = authorization_scheme::openid_connect,
-						.username = std::move(irods_name),
+						.username = *std::move(irods_username),
```
korydraughn, line 350: Shouldn't this be `std::move(*irods_username))`?

endpoints/authentication/src/main.cpp
```diff
@@ -581,7 +567,7 @@ namespace irods::http::handler
 							.get<int>();
 					auto bearer_token = irods::http::process_stash::insert(authenticated_client_info{
 						.auth_scheme = authorization_scheme::openid_connect,
-						.username = std::move(irods_name),
+						.username = *std::move(irods_username),
```
korydraughn, line 570: `std::move(*irods_username))`?

core/src/common.cpp
```diff
@@ -237,6 +385,30 @@ namespace irods::http
...
+					return {.response = fail(status_type::unauthorized)};
+				}
+
+				// Do mapping of user to irods user
+				auto user{map_json_to_user(json_res)};
+				if (user) {
+					return {.client_info = {.username = *std::move(user)}};
+				}
+
+				log::error("{}: Could not find a matching user.", __func__);
```
korydraughn, line 408: Is there anything worth we can/should include about the user to help the admin here?

core/src/common.cpp
```diff
@@ -207,6 +218,143 @@ namespace irods::http
...
+
+		const auto port{get_port_from_url(_url)};
+
+		req.set(beast::http::field::host, create_host_field(_url, *port));
+		req.set(beast::http::field::user_agent, irods::http::version::server_name);
+		req.set(beast::http::field::content_type,
+		        "application/x-www-form-urlencoded");
+		req.set(beast::http::field::accept, "application/json");
+
+		if (const auto& secret_key{irods::http::globals::oidc_configuration().find("client_secret")}; secret_key != std::end(irods::http::globals::oidc_configuration())) {
```
korydraughn, line 275: Drop `&`?

`secret_key` is an iterator.

core/src/common.cpp
```diff
@@ -207,6 +218,143 @@ namespace irods::http
...
+		auto res{tcp_stream->communicate(req)};
+		
+		log::debug("{}: Received the following response: [{}]", __func__, res.body());
+
+		// JSONize response
+		return nlohmann::json::parse(res.body());
+	}
+
+	auto map_json_to_user(const nlohmann::json& _json) -> std::optional<std::string> {
+		const auto oidc_config{irods::http::globals::oidc_configuration()};
```
korydraughn, line 327: Missing `&`?

core/src/common.cpp
```diff
@@ -207,6 +218,143 @@ namespace irods::http
...
+		const static bool use_user_claim{oidc_config.contains("irods_user_claim")};
+		const static bool use_attibute_mapping{oidc_config.contains("user_attribute_mapping")};
```
korydraughn, line 329: Consider making these iterators since the config is readonly.

Then the rest of the code can do a dereference without needing to do another lookup.

core/src/common.cpp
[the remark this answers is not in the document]
korydraughn: Please mark as resolved since the code has changed. I left new comments.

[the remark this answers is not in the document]
korydraughn: Nevermind, I'll go ahead and do it.

core/src/common.cpp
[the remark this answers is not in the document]
korydraughn: Got it. Marking as resolved since the code has changed and new comments have appeared.

core/src/main.cpp
MartinFlores751: Removed.

README.md
MartinFlores751: Added simple README items.

endpoints/authentication/src/main.cpp
MartinFlores751: `*std::move(irods_username)` 'steals' the value contained in the optional. Doing this the other way leaves an optional that claims to hold something, but the value is missing.

Doing it this way should make potential future errors more obvious.

endpoints/authentication/src/main.cpp
MartinFlores751: Words here: https://github.com/irods/irods_client_http_api/pull/252#discussion_r1513354505

README.md
```diff
@@ -263,9 +263,14 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
                 // The client id given to the application by OIDC provider.
                 "client_id": "irods_http_api",
 
-                // The client secret used for accessing the introspection endpoint
+                // The client secret used for accessing the introspection endpoint.
+                // Optional unless running as a protected resource.
                 "client_secret": "xxxxxxxxxxxxxxx",
 
+                // The OIDC mode the HTTP API will run as.
+                // The value has to be "client" or "protected_resource"
```
korydraughn, line 271: Missing period.

korydraughn, line 271: Please update the documentation to follow the same formatting/style as `/irods_client/tls/client_server_policy`.

List the supported options and give a brief description explaining what they do and/or how they change the interface.

README.md
```diff
@@ -274,9 +279,22 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
                 "state_timeout_in_seconds": 600,
 
                 // The name of the OIDC claim which provides the mapping of an
-                // OIDC user to an iRODS user account
+                // OIDC user to an iRODS user account.
+                // Only 'irods_user_claim' or 'user_attribute_mapping' can be
+                // declared at once.
```
korydraughn, line 284: Please update the documentation to follow the same formatting/style as `/irods_client/tls/client_server_policy`.

List the supported options and give a brief description explaining what they do and/or how they change the interface.

README.md
```diff
@@ -274,9 +279,22 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
...
                 "irods_user_claim": "irods_username",
 
+                // The mapping of user to the provided values. All values must
+                // be match to map an OIDC user to and iRODS user account.
```
korydraughn, line 288: `"All values must be match to map an OIDC user to and iRODS user account."`?

endpoints/authentication/src/main.cpp
korydraughn: But... that transfers ownership of the optional rather than the string, right?

Once we move the string, does the string in the optional matter?
How does `*std::move(irods_username)` result in the string inside of the optional moving into `.username`?

korydraughn: Found it! See comment in example on this page: https://en.cppreference.com/w/cpp/utility/optional/operator*

There may be other places in the code that may do it the way that I did. Please open an issue for verifying/updating the code to do it in the way you've specified.

core/src/common.cpp
MartinFlores751: We could say "check either irods_user_claim or user attribute mapping", but I think that might not be best?

Just because this may actually be a common error (i.e. non-irods/non-mapped user trying to access the endpoint).
Might be worth going from 'error' to 'warn'.

core/src/common.cpp
```diff
@@ -237,6 +385,30 @@ namespace irods::http
 		// Verify the bearer token is known to the server. If not, return an error.
 		auto mapped_value{irods::http::process_stash::find(bearer_token)};
 		if (!mapped_value.has_value()) {
+			// If we're running as a protected resource, assume we have a OIDC token
+			if (irods::http::globals::oidc_configuration().at("run_as_protected_resource").get<bool>()) {
```
korydraughn, line 389: `run_as_protected_resource`?

endpoints/authentication/src/main.cpp
MartinFlores751: Yep, the example on the operator* page is where I saw it.

>  How does *std::move(irods_username) result in the string inside of the optional moving into .username?

I'm pretty sure the idea is to create an 'rvalue' of the optional to move the data from the original to it. Then, you dereference the newly created rvalue to get the string, which should still be a move, since there's likely an rvalue optimization for it.

> There may be other places in the code that may do it the way that I did. Please open an issue for verifying/updating the code to do it in the way you've specified.

Will do.

core/src/common.cpp
korydraughn: I think I agree. The current message is good/fine.

How do you feel about logging a substring of the bearer token?

Also, logging as a warning sounds good.

core/src/common.cpp
MartinFlores751: Done.

endpoints/authentication/src/main.cpp
MartinFlores751: Done.

core/src/common.cpp
trel: what would the (partial) token tell the admin that is helpful?   the access token is time-bound / ephemeral, right?

doesn't feel good to log it on first glance...

korydraughn: We don't have to handle it now. We can change things later.

README.md
MartinFlores751: Resolving in favor of https://github.com/irods/irods_client_http_api/pull/252#discussion_r1513413216

README.md
MartinFlores751: Does this make sense in this case? claim & mapping are not "values", they're the actual "keys".

Is there a better format for this case?

korydraughn: It's more about making the docs consistent. All properties accepting a range of values have docs which list the acceptable values. Maintaining that is good IMO.

A better example of this is the log_level property.

> claim & mapping are not "values", they're the actual "keys".

It's not clear to me why that's important. Please elaborate.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

README.md
MartinFlores751: Done, check descriptions to see if like.

README.md
MartinFlores751: Brain-keyboard latency, should make sense now.

README.md
MartinFlores751: I think it would be confusing, since the format is used throughout for 'values'. Throwing it in for 'keys' would get you if you don't look closely at it.

Maybe we're talking a bit past each other here? Not sure if my 'keys' and 'values' are clear here.

core/src/common.cpp
```diff
@@ -325,16 +325,16 @@ namespace irods::http
...
 	auto map_json_to_user(const nlohmann::json& _json) -> std::optional<std::string> {
 		const auto& oidc_config{irods::http::globals::oidc_configuration()};
-		const static bool use_user_claim{oidc_config.contains("irods_user_claim")};
-		const static bool use_attibute_mapping{oidc_config.contains("user_attribute_mapping")};
+		const static auto user_claim{oidc_config.find("irods_user_claim")};
+		const static auto attribute_mapping{oidc_config.find("user_attribute_mapping")};
 
-		if (use_user_claim) {
-			if(auto claim{_json.find(oidc_config["irods_user_claim"].get_ref<const std::string&>())}; claim != std::end(_json)) {
+		if (user_claim == std::end(oidc_config)) {
```
korydraughn, line 331: Shouldn't this be `!=`?

core/src/common.cpp
```diff
@@ -325,16 +325,16 @@ namespace irods::http
...
+			if(auto claim{_json.find(user_claim->get_ref<const std::string&>())}; claim != std::end(_json)) {
 				return {*claim};
 			}
 		}
-		else if (use_attibute_mapping) {
-			const auto& mappings{oidc_config["user_attribute_mapping"]};
+		else if (attribute_mapping == std::end(oidc_config)) {
```
korydraughn, line 336: Shouldn't this be `!=`?

README.md
korydraughn: The docs for the property need to include all of the information for a user/admin to understand how to use it. One could argue that everything listed in the json file is a value. It depends on how you look at it. As long as the docs are clear, there's no issue.

```js
// Defines how user mapping is carried out.
//
// The following values are supported:
// - irods_username         
// - user_attribute_mapping
//
// When set to "irods_username", the HTTP API will ...
//
// When set to "user_attribute_mapping", the HTTP API will ...
"irods_user_claim": "irods_username",
```

That is just an example. The docs make it clear what is allowed and then proceeds to explain what the HTTP API does when set to either of the values.

Whether the value represents a key is irrelevant because that may not be true in the future.

Does that make sense or am I misunderstanding?

MartinFlores751: Still think we're not there yet.

We're talking about two separate 'keys', `irods_user_claim` and `user_attribute_mapping`:
```json
"irods_user_claim": "COOL_USERNAME_HERE",

"user_attribute_mapping": {
  "COOL_USERNAME_HERE": {
    "not": "important_right_now",
    "minor": "details_here"
}
```

Only one of them can be active at a time.

---

Mini discuss with Kory:

For the time being, we're just going to document the options individually, and briefly mention that the options cannot be used together.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

README.md
MartinFlores751: Done. Did a minor reword, let know if there's a better sentence.

MartinFlores751:
Rebased onto main. Still need to test.

MartinFlores751:
In `client` mode, all test pass.

In `protected_resource` mode, we get 9 tests failing:
```
...
FAILED test_auth_endpoint.py::test_oidc_resource_owner_password_credentials_login[bob-bob-200] - assert 400 == 200
FAILED test_auth_endpoint.py::test_oidc_resource_owner_password_credentials_login[--401] - assert 400 == 401
FAILED test_auth_endpoint.py::test_oidc_resource_owner_password_credentials_login[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb-401]
FAILED test_auth_endpoint.py::test_oidc_authorization_code_login - AssertionError: assert 405 == 200
FAILED test_auth_endpoint.py::test_oidc_authorization_code_non_irods_user - AssertionError: assert 405 == 200
FAILED test_auth_endpoint.py::test_get_oidc_errors[placeholder-bad_code-400] - assert 405 == 400
FAILED test_auth_endpoint.py::test_get_oidc_errors[-bad_code-400] - assert 405 == 400
FAILED test_auth_endpoint.py::test_get_oidc_errors[placeholder--400] - assert 405 == 400
FAILED test_auth_endpoint.py::test_get_oidc_errors[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb-400]
...
```

I would say these failures are expected, since we effectively disable the 'client' related items, and the tests are not made to handle this case yet. Not sure if these are the HTTP codes we want to return, though.

korydraughn:
I agree. Those tests seem specific to `client` mode and that's ok.

Consider adding new tests which target the `protected_resource` mode.

You can add options to the config.py file to ease some of the automation.

trel:
is there also a place to 'skip' tests in certain modes?     not sure the approach vs doubling up lines of code...

MartinFlores751:
> Consider adding new tests which target the `protected_resource` mode.

I'll try and cook up a test, hoping I don't have to make another client.

> is there also a place to 'skip' tests in certain modes? not sure the approach vs doubling up lines of code...

We can skip the tests, which I'm considering. If the test code for the modes varies a good bit, will likely make a new function to test the new mode.

trel:
>  If the test code for the modes varies a good bit, will likely make a new function to test the new mode.

that feels correct.

MartinFlores751:
> > Consider adding new tests which target the `protected_resource` mode.
> 
> I'll try and cook up a test, hoping I don't have to make another client.

Will likely need either a mini http server, or an OIDC client library to make a python test.
This is due to the part where the OpenID Provider returns a response to the 'redirect_uri', which we have to listen for.

> > is there also a place to 'skip' tests in certain modes? not sure the approach vs doubling up lines of code...
> 
> We can skip the tests, which I'm considering. If the test code for the modes varies a good bit, will likely make a new function to test the new mode.

Skipped the tests that were failing.

korydraughn:
Got it. Think about the simplest path forward to a 0.3.0 release and we'll discuss that tomorrow.

MartinFlores751:
All tests passed. Manual testing shows in client mode, access tokens are treated like iRODS bearer tokens (effectively rejected). In protected resource mode, access tokens that are expired error out. If the access token is not expired, but the token does not map to a user, it errors out.  Otherwise, it proceeds to do as requested, acting as the mapped user.

TLDR, scripted tests passed, and manual tests seem good too.

trel:
most excellent - thank you.

API.md
```diff
@@ -84,6 +89,18 @@ This will open a browser window that will allow you to authenticate in.
 ##### Response
 The bearer token should be returned and viewable in the browser window after authenticating.
 
+#### iRODS as an OAuth Protected Resource
+While running in this mode, the HTTP API does not provide any grants to which you can authenticate to.
```
trel, line 93: "authenticate to" -> "authenticate"

API.md
```diff
@@ -30,10 +30,15 @@ Reauthentication can be performed at anytime and will result in a brand new toke
...
 
-For authenticating with OpenID Connect, there are two methods:
+For authenticating with OpenID Connect, there are three methods, two of which are run as clients:
 
-- Resource Owner Password Credentials Grant
-- Authorization Code Grant
+- Client-Based
+  - Resource Owner Password Credentials Grant
+  - Authorization Code Grant
+- Protected Resource
```
korydraughn, line 38: You can link each of these to the sections that follow, allowing the reader to jump to the section of interest.

API.md
MartinFlores751: Done.

API.md
```diff
@@ -30,10 +30,15 @@ Reauthentication can be performed at anytime and will result in a brand new toke
...
+
+To run a client-based scheme, you must set `'mode'` in the `openid_connect` stanza to `'client'`.
+Otherwise, to run the HTTP API as an OAuth protected resource, set `'mode'` to `'protected_resource'`.
```
korydraughn, line 41: _This is motivated by my previous comment._

What if each section had a line that stated something like ...
```
To enable this auth flow, under the `openid_connect` stanza, set `mode` to `client`.
```

Then everything the admin needs for understanding how to make it work is right there within each section.

These lines would be removed, of course.

Thoughts?

API.md
```diff
@@ -84,6 +89,18 @@ This will open a browser window that will allow you to authenticate in.
...
 The bearer token should be returned and viewable in the browser window after authenticating.
 
+#### iRODS as an OAuth Protected Resource
+While running in this mode, the HTTP API does not provide any grants to which you can authenticate to.
+Instead, you must implement an OAuth client yourself, authenticate, and provide an access token as a
+Bearer token when querying the HTTP API endpoints.
+
+In order for the access token to be accepted, two conditions must be met:
+1) The OpenID Provider recognizes the token as valid and active
+2) The token must be able to be mapped to a user, using either `irods_user_claim` or `user_attribute_mapping`.
```
korydraughn, line 99: Missing a period for (1).

The use of parens caught my attention. Did you know it would render as a numbered list? I always thought you had to use "1.".

korydraughn (commented):
Looking really good.

API.md
MartinFlores751: You're right, I just defaulted to org lists instead.

trel: github markdown does either, yes.

API.md
MartinFlores751: Can do.

API.md
MartinFlores751: Done. Changed to use '.' version.

API.md
MartinFlores751: Done.

API.md
MartinFlores751: Made links, they should work.

README.md
```diff
@@ -271,9 +281,20 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
                 "state_timeout_in_seconds": 600,
 
                 // The name of the OIDC claim which provides the mapping of an
-                // OIDC user to an iRODS user account
+                // OIDC user to an iRODS user account.
+                // 'irods_user_claim' and 'user_attribute_mapping' cannot be used together
```
korydraughn, line 285: - Replace single quotes with double quotes for consistency
- Missing period

README.md
```diff
@@ -271,9 +281,20 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
...
                 "irods_user_claim": "irods_username",
 
+                // The mapping of a user to the provided values. All values must
+                // be matched to map an OIDC user to an iRODS user account.
+                // 'irods_user_claim' and 'user_attribute_mapping' cannot be used together
```
korydraughn, line 290: - Replace single quotes with double quotes for consistency
- Missing period

test/test_auth_endpoint.py
```diff
@@ -1,36 +1,46 @@
...
 import logging
 import html
 
-# Pragmatic
-# Search for IRODS_HTTP_API_BASE_URL
-# and IRODS_CLIENT_VERSION
-# TOOD: Consider turning this into a fixture
-AUTH_URL = 'http://127.0.0.1:9000/irods-http-api/0.2.0/authenticate'
+skip_protected_resource = pytest.mark.skipif(
+    test_config['openid_connect']['mode'] == 'protected_resource', reason='Not supported in "protected_resource" mode.')
```
korydraughn, line 10: What happens if the config options aren't defined?

test/test_auth_endpoint.py
```diff
@@ -1,36 +1,46 @@
...
 
-@pytest.mark.parametrize("username, password, expected_result", [('bob', 'bob', requests.codes.ok),
-                                                                 ('', '', requests.codes.unauthorized),
+skip_client = pytest.mark.skipif(
+    test_config['openid_connect']['mode'] == 'client', reason='Not supported in "client" mode.')
```
korydraughn, line 13: What happens if the config options aren't defined?

test/test_auth_endpoint.py
```diff
@@ -1,36 +1,46 @@
...
+
+@pytest.fixture
+def irods_http_api_url_base():
+    return f'http://{test_config["host"]}:{test_config["port"]}{test_config["url_base"]}'
```
korydraughn, line 17: Need to handle missing config options?

test/test_auth_endpoint.py
```diff
@@ -75,31 +86,33 @@ def test_oidc_authorization_code_non_irods_user():
...
 
-@pytest.mark.parametrize("state, code, expected_result", [('placeholder', 'bad_code', requests.codes.bad_request),
-                                                          ('', 'bad_code', requests.codes.bad_request),
-                                                          ('placeholder', '', requests.codes.bad_request),
-                                                          ('a'*200, 'b'*200, requests.codes.bad_request)])
-def test_get_oidc_errors(state, code, expected_result):
-    res = requests.get(AUTH_URL, params={'state': state, 'code': code})
-
-    assert res.status_code == expected_result
+    if test_config['openid_connect']['mode'] == 'client':
```
korydraughn, line 106: Probably should check for existence of properties?

test/test_auth_endpoint.py
```diff
@@ -1,36 +1,46 @@
...
                                                                  ('not', 'valid', requests.codes.bad_request),
-                                                                 ('a'*200, 'b'*200, requests.codes.unauthorized)])
-def test_oidc_resource_owner_password_credentials_login(username, password, expected_result):
+                                                                 pytest.param('a'*200, 'b'*200, requests.codes.unauthorized, marks=skip_protected_resource)])
+def test_oidc_resource_owner_password_credentials_login(username, password, expected_result, auth_url):
     to_be_encoded = f'{username}:{password}'
     encoded_user_pass = base64.b64encode(to_be_encoded.encode())
 
-    res = requests.post(AUTH_URL, headers={'Authorization': f'iRODS {encoded_user_pass.decode()}'})
+    res = requests.post(auth_url, headers={'Authorization': f'iRODS {encoded_user_pass.decode()}'})
```
korydraughn, line 31: Does `.decode()` default to utf-8?

korydraughn (commented):
Looks like this is almost ready.

Exciting stuff. Are we missing anything? Does the README contain everything the admin needs to know about the new features? Are there sharp corners that need more documentation?

test/test_auth_endpoint.py
MartinFlores751: Yes 👍

test/test_auth_endpoint.py
MartinFlores751: Can you be more specific? Do you mean the entire `test_config`? The `mode` option? Yes to all of the above?

test/test_auth_endpoint.py
korydraughn: Good.

test/test_auth_endpoint.py
korydraughn: If `openid_connect` and/or `mode` is not defined, this will result in a lookup error.

The test harness should handle missing options gracefully. Either by exiting with a good log message or falling back to a reasonable default.

For this specific case, logging a good error message and exiting seems like the correct approach.

Thoughts?

MartinFlores751: Would a schema make more sense? Validate early, fail otherwise?

MartinFlores751: I thought the tests didn't have a schema. We're talking about `config.py` not the server, which has its own config. But yes, the idea is to catch and give a stern talking.

korydraughn: > we... already have a schema? so at this point, we should just... catch a lookup error and exit angrily?

This is for test's config file (which is just a python file).

> Would a schema make more sense? Validate early, fail otherwise?

That would work. How would you do that with a python file though? Unless you're talking about changing the file to a json file.

trel: ah, okay.    carry on with the discussion.

if not present - exit:
 - pro: easier, less confusing to admins, finds errors quickly
 - con: does not allow for easy upgrades if/when we introduce a newly required setting/value

if not present - use default:
 - pro: easier upgrades, more forgiving to admins who didn't get it right
 - con: could be surprising to find things are running with a 'magic' default

i'm leaning to exit early on this one...

korydraughn: I'm in the middle on this one.

For some things, it's clear they need to be required. For others, a reasonable default makes sense. For example, the OIDC stuff is required for the OIDC tests, but can be ignored for the non-OIDC tests.

README.md

README.md

test/test_auth_endpoint.py
MartinFlores751: I pushed up an example schema. Worked when I ran it.

korydraughn: I see the call to `validate()` at the bottom of the python file. When is that run?

Perhaps each test suite needs its own schema?

MartinFlores751: If recall correctly, it's run on import.

> Perhaps each test suite needs its own schema?

You mean besides the HTTP API?

korydraughn: No. There are two test suites in this repo. The OIDC tests and the non-OIDC tests.

They have different requirements, therefore it may be worth using different schemas. That will make it clear which options are absolutely necessary for each suite.

If we don't like two schemas, then we can say all options are required and include comments in the config.py explaining what each option is for.

MartinFlores751: Think the two schemas approach was too awkward, given that only the 'openid_connect' option is sectioned out for easy disabling. If we want a multiple schema approach, it would be better to section off the 'required' options and the 'non-oidc' test options.

korydraughn: Or... we just make everything required. Then the person running the tests has to make sure everything is configured correctly. No need for anything fancy.

We just need to include good documentation for testing.

That keeps things simple and straight forward.

Thoughts?

trel: just saying this out loud... if they're required (rather than present/not), then we have to give them a way to 'turn it off'.

korydraughn: Very true. The config.py currently favors the non-oidc tests.

I'm just throwing out ideas here. Who knows if they're good or bad :-)

test/config.py
```diff
@@ -20,5 +25,73 @@
...
+                'password': {
+                    'type': 'string'
+                }
+            },
+            'required': [ 'username', 'password' ]
+        }
+    }
 }
+
+validate(instance=test_config, schema=schema)
```
korydraughn, line 97: Can `validate` collide with things in the test python scripts?
Is it safer to qualify it with `jsonschema`?

korydraughn (commented):
Changes look good.

Do the OIDC and non-OIDC tests pass?

test/config.py
MartinFlores751: Within the import/config.py itself, no. Does `validate(...)` get out? Yes, if you use `config.validate(...)` after an  `import config`, or one of the other various imports to pull in validate (e.g., `from config import validate`). I think it may collide if you do a `from config import *`, but that's bad practice anywho.

korydraughn: So if we used the following, that would remove the possibility of name collisions, right?
```python
import jsonschema
# ... other code ...
jsonschema.validate(args...)
```

MartinFlores751: It turns into `config.jsonschema.validate`. Can change it to `import jsonschema` if you want, no real preferences here.

korydraughn: Hmm, so we can't keep the call to `validate` private to config.py?

If that's not possible, then what you have is fine.

MartinFlores751: I think that would require some module magic/organization that I'm not too familiar with. Maybe when the config grows bigger it could be something we look into.

korydraughn: Agreed. Not a blocker right now.

korydraughn:
Please resolve the remaining comments and squash to taste if everything is working as expected.

test/test_auth_endpoint.py
MartinFlores751: We decided to make all the properties required for the moment.

test/test_auth_endpoint.py
MartinFlores751: Resolved with schema.

test/test_auth_endpoint.py
MartinFlores751: Resolved with schema.

test/test_auth_endpoint.py
MartinFlores751: Resolved with schema.

MartinFlores751:
Squashed.

trel:
please order the '108, 155, 235' in the commit messages.

README.md
```diff
@@ -271,9 +281,20 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
...
+                // "irods_user_claim" and "user_attribute_mapping" cannot be used together.
                 "irods_user_claim": "irods_username",
 
+                // The mapping of a user to the provided values. All values must
+                // be matched to map an OIDC user to an iRODS user account.
+                // "irods_user_claim" and "user_attribute_mapping" cannot be used together.
+                "user_attribute_mapping": {
+                    "irods_username": {
+                        "sub": "123-abc-456-xyz",
+                        "email": "rods_user@email.example"
```
trel, line 294: could use example.org as the domain

API.md
```diff
@@ -84,6 +90,20 @@ This will open a browser window that will allow you to authenticate in.
 ##### Response
 The bearer token should be returned and viewable in the browser window after authenticating.
 
+#### iRODS as an OAuth Protected Resource
+To run as an OAuth protected resource, set `mode` to `protected_resource`.
```
trel, line 94: let's make this sentence look more similar to the above...

"set `mode` in the `openid_connect` stanza to `protected_resource `."

README.md
MartinFlores751: Just to be sure, prefer `example.org` in the docs, right?

trel: right, this is the README.md, yes?

API.md
MartinFlores751: Done.

README.md
MartinFlores751: Changed it!

korydraughn (commented):
Looks good to me.

Let's get one more approval before squashing, no #'s.

korydraughn:
I was too slow. No problem.

korydraughn (approved):
Pound it once @trel approves it.

Good stuff.

trel:
hooray - let's do it.

MartinFlores751:
Pounded.

trel:
Please close any issues mentioned.

Issue #227, opened 2024-01-23
OIDC Configuration Parsing Error and Key Mismatch in iRODS HTTP API


bl000m:
## Feature

- Please describe the feature you are requesting and your use case.

# OIDC for Single Sign-On with Indigo IAM OIDC Provider

Both myself and @sigau are encountering issues while configuring the OIDC settings. The config.json has recently changed, and if we retain the previous version:
```
 "http_server": {
        "host": "0.0.0.0",
        "port": 9000,
        "log_level": "trace",
        "authentication": {
            "eviction_check_interval_in_seconds": 60,
            "basic": {
                "timeout_in_seconds": 3600
            },
            "oidc": {
                "config_host": "fits-indigo-iam-test.in2p3.fr",
                "port": 8080,
                "well_known_uri": "https://fits-indigo-iam-test.in2p3.fr/.well-known/openid-configuration",
                "client_id": "***",
                "redirect_uri": "http://***:5599/irods-http-api/0.1.0/authenticate",
                "state_timeout_in_seconds": 600
            }
        },
```
Executing `docker run --rm --name irods_http_api -v ./config.json:/config.json:ro -p 9000:9000 irods/irods_http_api` results in the following output, where the HTML of the OIDC provider login page is parsed as if it were the expected JSON:
```
[2024-01-23 13:26:22.860] [P:1] [info] [T:1] Initializing server.
[2024-01-23 13:26:22.860] [P:1] [trace] [T:1] Verifing OIDC endpoint configuration
[2024-01-23 13:26:22.878] [P:1] [debug] [T:1] Got the following back: 

<!DOCTYPE html>
<html lang="en_US">

<head>

... (rest of the HTML content)
</html>

Error: [json.exception.parse_error.101] parse error at line 11, column 1: syntax error while parsing value - invalid literal; last read: '<U+000A><U+000A><U+000A><U+000A><U+000A><U+000A><U+000A><U+000A><U+000A><U+000A><'
```
- Even with the new config.json configuration, replacing oidc with openid_connect:

```
"openid_connect": {
                "provider_url": "https://fits-indigo-iam-test.in2p3.fr:8080/.well-known/openid-configuration",
		"client_id": "***",
                "redirect_uri": "http://***:5599/irods-http/0.1.0/authenticate",
                "state_timeout_in_seconds": 600
            }
```
We encounter the following issue:
```
Initializing server.
[...]

Issue #248, opened 2024-02-06
Allow customization of Access Token Scope


MartinFlores751:
We should allow the user to specify additional scopes along with the default `oidc` scope for further flexibility.

Example (Implementation may be different):
```json
"oidc": {
  "additional_scopes": ["irods", "custom", "misc"]
}
```

MartinFlores751:
Perhaps even allow the `oidc` scope to be optional?

korydraughn:
We haven't spent any time thinking through the details due to other priorities, but we'll address this once work on 0.4.0 begins.

korydraughn:
> By validating scopes (a scope mapping could be added in the conf to establish a relation between scope names and API ops), the resource server will expose only those API endpoints compatible with them, increasing security and providing a way of restricting API access dynamically.

True. iRODS already has a permission model though. Admins can already block access to endpoints (e.g. /data-objects) through the use of a proxy (e.g. nginx or httpd). While we can certainly improve the use of scopes, it's difficult to determine whether it'll prove to be useful to people.

That's just my initial thoughts.

We haven't made any decisions regarding this yet. We're very much open to hearing use-cases from people. Feel free to post your thoughts and ideas here.

korydraughn:
Thanks. This is good feedback. We'll keep this in mind as we approach the next release.

If you think of anything else, please post it here.

PR #253, opened 2024-02-13
Access token handling

bl000m:
The Access Token Validation feature introduces an endpoint for validating access tokens, ensuring its integrity and authenticity  before granting access to protected resources.

I experimented with your Api in a test instance to allow a portal (still a work in progress) to access iRODS through the API using an access token issued by the same OIDC provider configured in the iRODS HTTP API (SSO), as discussed [here](https://github.com/irods/irods_client_http_api/issues/227). 

I implemented this feature to test OIDC compatibility, and everything worked seamlessly on my end.

I thought it would be interesting for you to test on your side and potentially integrate it if you believe it could benefit other users as well. 
Below some details:

**Endpoint**
The` /validate-token` endpoint is responsible for validating the provided access token and generating a new bearer token.

_Request_
`curl -X POST -H 'Authorization: Bearer <access_token>' http://localhost:<port>/irods-http-api/<version>/validate-token`

_Response_
If the access token is valid, the response will contain a new bearer token, which can be used for subsequent operations.
```
{
    "success": true,
    "bearer_token": "<new_bearer_token>"
}
```
This endpoint follows the OAuth 2.0 protocol for token validation and issuance.

**Implementation Details**

**Testing**: Conducted thorough testing on a test instance running on VM with Ubuntu 22.04, including setup script for launching Docker container and updating dependencies  => `launch_docker.sh`
**Dependency Management**: Updated the CMake architecture of dependencies to ensure compatibility.
**Integration**: Integrated the endpoint to create, if the access token is validated, the bearer token that matches the `struct `currently used in the `resolve_client_identity` function to cast the info extracted from the token (struct `authenticated_client_info`).
**Portal Integration**: Implemented functionality to allow the portal to access iRODS by 
- making a POST Request to the iRODS HTTP API `/authenticate` endpoint to obtain the` authorization code`
- using this code to obtain the `AT` from the OIDC provider 
- finally making a POST Request to the newly implemented iRODS HTTP API `/validate-token` endpoint to obtain the iRODS Bearer Token to be used for further fetching

**Scope Enhancement:** Added the `"profile"` scope in the implementation of the code for the `/authenticate` handler in the iRODS HTTP API (in the args of the request url building) to ensure the correct authorization code is obtained from the OIDC provider, allowing access to the `"preferred_username`" claim in the access token.

**Documentation Updates**
- Updated the README and API documentation to reflect the new feature
- Added comments within the codebase for clarity.

**Note**
This feature aims to ensure compatibility with OIDC providers and enable seamless integration with portals and other external systems. However, further testing with different OIDC providers may be required to confirm compatibility.


korydraughn:
Thank you for this.

We'll review and let you know our thoughts.

MartinFlores751:
Hello @bl000m, we were hoping to get your opinion on PR #252. The work done is similar to what you have done, and we would like to know if it would solve your particular problem too.

bl000m:
hello @MartinFlores751 

Wth pleasure:

- The program is not compiling because "run_as_protected_resources" is set as type string in the JSON schema in main.cpp. It needs to be boolean since you retrieve its value here as boolean; otherwise, the program will not compile:
```
if (oidc_stanza_exists) {
    static const auto is_client{irods::http::globals::oidc_configuration().at("run_as_protected_resource").get<bool>()};
    return is_client;
}
```
So:
```
// main.cpp
"run_as_protected_resource": {{
                                    - "type": "string"
                                    + "type": "boolean"
                                }},
```
and

```
- "run_as_protected_resource": "<string>",
+ "run_as_protected_resource": "<boolean>",
```
Also, when you: 
```
const auto& mappings{irods::http::globals::oidc_configuration().at("user_attribute_mapping")};, 
```
you are referring to something that is not required in the JSON schema and that throws an error. So either you change the JSON schema in main.cpp, adding the `user_attribute_mapping`, or you remove the section and refer just to the `irods_user_claim` value as mapping with the OIDC provider user, that in my opinion is the best solution. 

Changing the `run_as_protected_resources`type to boolean and removing the  `user_attribute_mapping`retrieving code snippet and replacing it with the `irods_user_claim`value as mapping value  I was able to make it work by first making a GET request to retrieve the authorization code, adding the 'profile' scope, and asking for the access token. Then, making a POST request to other endpoints, in particular, the collection list.

My thoughts:

In your code, you've opted to enable fetching endpoints with an OIDC access token. However, this requires altering the logic in `resolve_client_identity` within `common.cpp`. Additionally, the iteration on `user_attribute_mapping`, even if it were correctly required, doesn't seem to be the optimal solution.

The `/authenticate` handler has grown to over 300 lines, with an abundance of conditional statements that hinder readability. While the choice of architecture is ultimately yours, I'd suggest considering alternatives to avoid overloading the `/authenticate` handler. Moreover, refraining from making significant changes to the logic in `resolve_client_identity` and minimizing the number of configuration parameters in `config.json` can prevent confusion.

A potential solution could be to create a new endpoint, as I proposed in my PR, dedicated to hitting, as you better did, the introspection OIDC provider endpoint for validation. This approach would entail sending back the iRODS token as a response to the validation request, that at the end of the day appears to be the best solution to grant access to fetch all endpoints. This alternative could offer a lighter solution, enhancing readability and promoting separation of concerns.

trel:
Additional scopes is https://github.com/irods/irods_client_http_api/issues/248

MartinFlores751:
Thank you for the comments, certainly brought up some good points.

> The program is not compiling because run_as_protected_resources is set as type string in the JSON schema in core/src/main.cpp.

> the user_attribute_mapping you are referring to is not required in the JSON schema and that throws an error.

Yes, these are mistakes that I will be fixing.

> ...just to the irods_user_claim value as mapping with the OIDC provider user, that in my opinion is the best solution, since the user_attribute_mapping logic seems a bit unintuitive, and I'm not sure it could work in all cases, for sure not in my case.

The idea is that the `user_attribute_mapping` is a more advanced version of `irods_user_claim`. The `irods_user_claim` wants the username directly, whereas the `user_attribute_mapping` is looking for several properties to match against. The reasoning behind this is that we assume not everyone has access to modify their OpenID Provider. As such, an alternate user mapping method is required. You are right that we should consider allowing `irods_user_claim` to work there too.

> The /authenticate handler signature function has grown to over 300 lines, with an abundance of conditional statements that hinder readability. While the architecture choice is ultimately yours, and I understand the strategy of consolidating various scenarios into a few main endpoint handlers to prevent dispersion, I recommend exploring alternatives to prevent overloading the /authenticate handler.

I understand the concerns about the `/authenticate` endpoint. We are planning to refactor the code to make it easier to follow.

> Moreover, refraining from making significant changes to the logic in resolve_client_identity and minimizing the number of configuration parameters in config.json can prevent confusion.

The new code in `resolve_client_identity` will be refactored. This should improve readability. On the configuration parameters, you are right. Reducing configuration parameters, if possible, would be beneficial.

> Creating a new endpoint for handling access token validation in order to release the iRODS token needed for accessing all endpoints could be beneficial, moving the hit_intro_endpoint logic into it. This approach would entail sending back the iRODS token as a response to the validation request, that at the end of the day appears to be the best solution to grant access to fetch all endpoints without revolutioning the existing code. This alternative could offer a lighter solution, enhancing readability and promoting separation of concerns.

I agree, using the iRODS token would be the lighter solution, however, the goal of #252 is to provide improved and easier integration with OpenID clients. By using the iRODS token, integration would be more difficult, as clients now need to handle an additional token instead of using what they already have. There are also additional benefits we receive by using access tokens instead of iRODS tokens, such as expiration and revocation handled by the OpenID Provider.

> ...it is necessary to include the profile scope in the arguments of the redirect URL to the authorization endpoint. ...

Yes, this will be resolved in #248.

PR #268, opened 2024-03-15
prepare for 0.3.0

trel:
building and testing now...


trel:
remaining ones to consider...

can these be generated/filled from a known variable?

```
./core/src/main.cpp:188:    "$id": "https://schemas.irods.org/irods-http-api/0.3.0/schema.json",
./test/config.py:9:    'url_base': '/irods-http-api/0.3.0',
./test/config.py:33:    '$id': 'https://schemas.irods.org/irods-http-api/test/0.3.0/test-schema.json',
```

korydraughn:
`main.cpp` - I believe this is doable. I can investigate for the next release. For 0.3.0, I think we leave it for now.

`config.py` - Yes, but it may be awkward to do. The best thing to do is to make it use `<version>` as well. Then it's on the dev to make sure they use the correct version number.

korydraughn:
The CMakeLists.txt defines macros which are used in main.cpp. The version is likely already available to main.cpp, we just have to tweak the code.

trel:
got it.  i'll leave these three for now, and we'll tackle post-0.3.0.

trel:
merging.     and one more build/test cycle with the real SHA before pushing to docker hub.

korydraughn:
Sounds good.

korydraughn (approved):

trel:
Captured in https://github.com/irods/irods_client_http_api/issues/269

Issue #133, opened 2023-11-01
Add operation to /resources endpoint that allows different properties of a resource to be modified


korydraughn:
- [x] main
- [x] 4-3-stable

This would provide functionality like `iadmin modresc`.

PR #279, opened 2024-07-01
[#133] Add support for modifying properties of a resource (main)

korydraughn:


endpoints/resources/src/main.cpp
```diff
@@ -238,17 +238,103 @@ namespace
...
+				if (value_iter == std::end(_args)) {
+					logging::error("{}: Missing [value] parameter.", fn);
+					return _sess_ptr->send(irods::http::fail(http::status::bad_request));
+				}
+
+				auto conn = irods::get_connection(client_info.username);
+
+				if (property_iter->second == "name") {
+					// TODO(???): Remove this once the resource administration library grows support
+					// for renaming resources.
```
korydraughn, line 263: Create issue for this.

endpoints/resources/src/main.cpp
```diff
@@ -221,49 +222,133 @@ namespace
...
+					input.arg0 = "modify";
+					input.arg1 = "resource";
+					input.arg2 = name_iter->second.c_str();
+					input.arg3 = "name";
+					input.arg4 = value_iter->second.c_str();
+
+					if (const auto ec = rcGeneralAdmin(static_cast<RcComm*>(conn), &input); ec < 0) {
+						constexpr const char* msg = "Could not change name from [{}] to [{}] for resource [{}]";
+						// NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+						THROW(ec, fmt::format(msg, input.arg3, input.arg4, input.arg2));
```
alanking, line 281: Feels like this message needs to be reworded or something. Currently, I think it will print something like this if we are attempting to rename a resource from resc1 to resc2:

"Could not change name from [name] to [resc2] for resource [resc1]."

An initial suggestion:

"Could not change property [name] to [resc2] for resource [resc1]."

...or something like that.

endpoints/resources/src/main.cpp
```diff
@@ -221,49 +222,133 @@ namespace
...
+					// TODO(???): Remove this once the resource administration library grows support
+					// for renaming resources.
+					//
+					// The following is equivalent to what the resource administration library would
+					// provide.
+					GeneralAdminInput input{};
+					input.arg0 = "modify";
+					input.arg1 = "resource";
+					input.arg2 = name_iter->second.c_str();
+					input.arg3 = "name";
```
alanking, line 275: At the risk of being less clear... this could be `property_iter->second.c_str()` instead of just "name".

endpoints/resources/src/main.cpp
```diff
@@ -221,49 +222,133 @@ namespace
...
+
+					if (value_iter->second == "up") {
+						status_enum = adm::resource_status::up;
+					}
+					else if (value_iter->second == "down") {
+						status_enum = adm::resource_status::down;
+					}
+					else {
+						logging::error(
+							"{}: Invalid value for [value] parameter. Received [{}].", fn, value_iter->second);
```
alanking, line 305: Should we state the name of the property ("status") in the message as well?

test/test_irods_http_api.py
```diff
@@ -2653,6 +2653,74 @@ def test_modifying_metadata_atomically(self):
...
+
+        try:
+            # The list of updates to apply in sequence.
+            property_map = [
+                ('name',        'test_modifying_resource_properties_renamed'),
+                ('type',        'passthru'),
+                ('host',        'example.org'),
+                ('vault_path',  '/tmp/test_modifying_resource_properties_vault'),
+                ('status',      'down'),
+                ('status',      'up'),
```
alanking, line 2678: Might be worth adding a test with an invalid value here since these are being checked in the HTTP API. Thoughts?

test/test_irods_http_api.py
```diff
@@ -2653,6 +2653,74 @@ def test_modifying_metadata_atomically(self):
...
+                        'op': 'modify',
+                        'name': resource,
+                        'property': p,
+                        'value': v
+                    })
+                    self.logger.debug(r.content)
+                    self.assertEqual(r.status_code, 200)
+                    self.assertEqual(r.json()['irods_response']['status_code'], 0)
+
+                    # Make sure to update the "resource" variable following a succesful rename.
```
alanking, line 2700:
```suggestion
                    # Make sure to update the "resource" variable following a successful rename.
```

alanking (commented):
Mostly noticed minor things

endpoints/resources/src/main.cpp
korydraughn: Good catch. Will need to adjust it in irods/irods as well.

endpoints/resources/src/main.cpp
korydraughn: It could, but I don't see much value in doing that for this case. Using the string literal removes any reliance on the compiler optimizing away potential memory accesses.

This is also identical to what will land in the resource administration lib.

I'm going to leave this as is for now.

endpoints/resources/src/main.cpp
korydraughn: Created #284 for this.

endpoints/resources/src/main.cpp
korydraughn: Will adjust.

test/test_irods_http_api.py
korydraughn: Done

test/test_irods_http_api.py
korydraughn: Done.

endpoints/resources/src/main.cpp
korydraughn: Done.

endpoints/resources/src/main.cpp
korydraughn: Done.

Created irods/irods#7877 too.

endpoints/resources/src/main.cpp
korydraughn: Done.

korydraughn:
Any last thoughts before I squash?

endpoints/resources/src/main.cpp
```diff
@@ -516,6 +604,14 @@ namespace
 				else if (const auto resc = adm::client::resource_info(conn, name_iter->second); resc) {
 					exists = true;
 
+					std::string_view status = "?";
```
alanking, line 607: The other member of the `resource_status` enumeration is `unknown`. Would it make more sense to default to "unknown" instead of "?"?

endpoints/resources/src/main.cpp
```diff
@@ -221,49 +222,136 @@ namespace
...
+					input.arg0 = "modify";
+					input.arg1 = "resource";
+					input.arg2 = name_iter->second.c_str();
+					input.arg3 = "name";
+					input.arg4 = value_iter->second.c_str();
+
+					if (const auto ec = rcGeneralAdmin(static_cast<RcComm*>(conn), &input); ec < 0) {
+						constexpr const char* msg = "Could not change property [name] to [{}] for resource [{}]";
+						// NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+						THROW(ec, fmt::format(msg, input.arg4, input.arg4, input.arg2));
```
alanking, line 281: `msg` only has 2 spots for substitutions. I think we need to remove one of these `input.arg4`s?

endpoints/resources/src/main.cpp
trel: question mark makes me... twitchy as well.

endpoints/resources/src/main.cpp
trel: yes

endpoints/resources/src/main.cpp
korydraughn: iRODS allows admins to set the status to any string AFAIK. From the help text of `iadmin modresc`.
```
Setting the resource status to 'down' will cause iRODS to ignore that
resource and bypass communications with that server.  'up' or other strings
without 'down' in them will restore use of the resource.  'auto' will allow
the Resource Monitoring System (if running) to set the resource status
to 'auto-up' or 'auto-down'.
```
Based on that text, the question mark should be replaced with `up`. The resource administration lib may need a tweak to better handle the status too. Perhaps it shouldn't interpret the status string. I'll review that implementation and see what it does exactly.

I'll need to think through whether `unknown` works.

korydraughn: Here's another thing to consider - should modifying the status be exposed at all?

endpoints/resources/src/main.cpp
korydraughn: Done.

endpoints/resources/src/main.cpp
alanking: That's a good question... And if it is exposed, do we support "any string that doesn't contain 'down'" as "up"? Do we support "auto"?

trel: Yes, the question mark is not a problem, just feels weird.   'unknown' captures that sense better than a symbol, in my opinion.

> should modifying the status be exposed at all? 

Administrators could be using that field for any number of things.   We should not remove it.

API.md
```diff
@@ -1394,6 +1394,49 @@ If an HTTP status code of 200 is returned, the body of the response will contain
...
+- type
+- host
+- vault_path
+- context
+- status
+- free_space
+- comments
+- information
+
+For the `status` property, `value` must be set to `up` or `down`. Attempting to pass a value other the ones described will result in an HTTP error.
```
trel, line 1423: This does not feel correct.   iRODS does not limit this value to 'up' or 'down'.

endpoints/resources/src/main.cpp
korydraughn: If we use "unknown", that means the resource could be up or down. That's my only hesitation about using that.

With that said, seeing "unknown" could make the admin reconsider the values they pass for the status when using tools other than the HTTP API.

API.md
alanking: Right, but the administration library only allows for `up` or `down` (https://github.com/irods/irods/blob/5e0e0c8dd5534b337cbced8a4c6ebbb39d363e51/lib/administration/resource/include/irods/resource_administration.hpp#L77-L83).

endpoints/resources/src/main.cpp
korydraughn: After some discussion, we've decided to return the resource status as is.

The HTTP API will continue to constrain the acceptable values for resource status to up and down.

korydraughn: Done.

test/test_irods_http_api.py
```diff
@@ -2714,6 +2714,129 @@ def test_modifying_metadata_atomically(self):
...
+
+        try:
+            # The list of updates to apply in sequence.
+            property_map = [
+                ('name',        'test_modifying_resource_properties_renamed'),
+                ('type',        'passthru'),
+                ('host',        'example.org'),
+                ('vault_path',  '/tmp/test_modifying_resource_properties_vault'),
+                ('status',      'down'),
+                ('status',      'up'),
```
korydraughn, line 2739: Add one or two more cases for status.

korydraughn: Wait, the HTTP API doesn't allow values other than these.

Resolving.

korydraughn:
Merging.

Issue #276, opened 2024-06-18
Add operations to manage specific queries


farbs03:
- [x] main
- [x] 4-3-stable
---
This would provide functionality like:
- `iadmin asq`
- `iadmin rsq`

This functionality is needed by the zone management tool.

PR #281, opened 2024-07-01
[#276] Add support for adding/removing specific queries (main)

korydraughn:


test/test_irods_http_api.py
```diff
@@ -2342,6 +2342,63 @@ def test_support_for_specific_queries(self):
...
+
+        finally:
+            # Show that rodsusers are NOT allowed to remove specific queries.
+            r = requests.post(self.url_endpoint, headers=rodsuser_headers, data={
+                'op': 'remove_specific_query',
+                'name': specific_query_name
+            })
+            self.logger.debug(r.content)
+            self.assertEqual(r.status_code, 200)
+            self.assertEqual(r.json()['irods_response']['status_code'], irods_error_codes.SYS_NO_API_PRIV)
```
alanking, line 2391: Should this be in the `finally` block? It rather feels like an additional test which should come at the end of the `try`.

korydraughn: Yes, agreed. Will update.

API.md
```diff
@@ -1275,6 +1275,69 @@ curl http://localhost:<port>/irods-http-api/<version>/query \
...
+{
+    "irods_response": {
+        "status_code": 0
+        "status_message": "string" // Optional
+    },
+    "rows": [
+        ["string", "string", "string"],
+
+        // Additional rows ...
+    ]
```
alanking, line 1306: I see that these are in the other responses, but it's not clear to me what this will contain. Is it just the rows that were updated?

API.md
```diff
@@ -1275,6 +1275,69 @@ curl http://localhost:<port>/irods-http-api/<version>/query \
...
+}
+```
+
+### remove_specific_query
+
+Removes a specific query.
```
alanking, line 1312: Would it be worth saying "SpecificQuery" or making reference to `iadmin rsq` or something? The phrase "specific query" reads to me like "I would like to remove this particular query" rather than the iRODS concept of a "Specific Query" (which is named such for parity with a "General Query" or GenQuery). Hope that makes sense... Maybe it's clear as it is.

alanking (commented):
Couple questions in the docs

API.md
korydraughn: Good question. Rephrasing that to use SpecificQuery or something is better. Will change.

API.md
korydraughn: That's a copy-paste issue. Will remove.

korydraughn: Done

API.md

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
#'d

Issue #300, opened 2024-07-11
Server segfaults when OIDC stanza and Bearer token are missing when user attempts to execute endpoint operation


korydraughn:
## Bug Report

The reason for the segfault is because the following code assumes the OIDC config is always present.
https://github.com/irods/irods_client_http_api/blob/fa7b44a264e88a5c92ad25de8bbb136d30fbd605/core/src/common.cpp#L402-L406

PR #304, opened 2024-07-11
[#300] Check OIDC configuration exists during bearer token resolution (main)

korydraughn:
No test for this since it requires the test to control the HTTP API server.

All tests pass. This did not include testing OIDC.

### DIFF
--- a/core/src/common.cpp
+++ b/core/src/common.cpp
@@ -401,9 +401,21 @@ namespace irods::http
 		// Verify the bearer token is known to the server. If not, return an error.
 		auto mapped_value{irods::http::process_stash::find(bearer_token)};
 		if (!mapped_value.has_value()) {
+			const auto& config = irods::http::globals::configuration();
+
+			// It's possible that the admin didn't include the OIDC configuration stanza.
+			// This use-case is allowed, therefore we check for the OIDC configuration before
+			// attempting to access it. Without this logic, the server would crash.
+			const auto oidc_config_iter =
+				config.find(nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"});
+			if (oidc_config_iter == std::end(config)) {
+				logging::debug("{}: No 'openid_connect' stanza found in server configuration.", __func__);
+				logging::error("{}: Could not find bearer token matching [{}].", __func__, bearer_token);
+				return {.response = fail(status_type::unauthorized)};
+			}
+
 			// If we're running as a protected resource, assume we have a OIDC token
-			if (irods::http::globals::oidc_configuration().at("mode").get_ref<const std::string&>() ==
-			    "protected_resource") {
+			if (oidc_config_iter->at("mode").get_ref<const std::string&>() == "protected_resource") {
 				body_arguments args{{"token", bearer_token}, {"token_type_hint", "access_token"}};
 
 				auto json_res{hit_introspection_endpoint(url_encode_body(args))};


trel:
looks good by inspection.

trel:
@MartinFlores751 please eyeball as well.

core/src/common.cpp
```diff
@@ -401,9 +401,20 @@ namespace irods::http
...
 		if (!mapped_value.has_value()) {
+			const auto& config = irods::http::globals::configuration();
+
+			// It's possible that the admin didn't include the OIDC configuration stanza.
+			// This use-case is allowed, therefore we check for the OIDC configuration before
+			// attempting to access it. Without this logic, the server would crash.
+			const auto oidc_config_iter =
+				config.find(nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"});
+			if (oidc_config_iter == std::end(config)) {
+				logging::trace("{}: No openid_connect configuration found in server configuration.", __func__);
```
alanking, line 412: I'm thinking `debug` may be slightly better here, but I'm also fine with `trace`.

alanking (approved):
Seems good to me, but definitely wait for @MartinFlores751 to have a look

core/src/common.cpp
korydraughn: I don't mind using debug. Will update.

MartinFlores751 (commented):
Change is fine as is. Rewording the log message might help make things a bit clearer.

core/src/common.cpp
```diff
@@ -401,9 +401,20 @@ namespace irods::http
...
 		auto mapped_value{irods::http::process_stash::find(bearer_token)};
 		if (!mapped_value.has_value()) {
+			const auto& config = irods::http::globals::configuration();
+
+			// It's possible that the admin didn't include the OIDC configuration stanza.
+			// This use-case is allowed, therefore we check for the OIDC configuration before
+			// attempting to access it. Without this logic, the server would crash.
+			const auto oidc_config_iter =
+				config.find(nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"});
+			if (oidc_config_iter == std::end(config)) {
```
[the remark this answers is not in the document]
korydraughn: Leaving as is for now.

We can refactor later if needed.

core/src/common.cpp
```diff
@@ -401,9 +401,20 @@ namespace irods::http
...
+				logging::debug("{}: No openid_connect configuration found in server configuration.", __func__);
```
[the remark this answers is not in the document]
korydraughn: done

korydraughn:
Tests still pass.

korydraughn:
I think we're ready for squashing.

Will wait for signal.

core/src/common.cpp
```diff
@@ -409,7 +409,8 @@ namespace irods::http
 			const auto oidc_config_iter =
 				config.find(nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"});
 			if (oidc_config_iter == std::end(config)) {
-				logging::debug("{}: No openid_connect configuration found in server configuration.", __func__);
+				logging::debug("{}: No 'openid_connect' stanza found in server configuration.", __func__);
+				logging::error("{}: Could not find bearer token matching [{}].", __func__, bearer_token);
```
alanking, line 413: Is this new error message intentional?

korydraughn: Yes. As a result of the oidc config not being defined, the bearer token cannot be found/matched.

alanking: Oh, I see. Is there any value in keeping the `debug` message separate, then? Seems like having both in the same message would be just as well. But I'm not knowledgeable in this area, so it may be useful to keep them separated.

korydraughn: Either way would be fine. I'm going to leave it as is and see how things turn out.

We can always adjust them later.

trel:
agree with the double logging message feeling weird... but good enough for now.

please squash.

korydraughn:
Squashed.

alanking (approved):
Seems ready. I say # it

korydraughn:
#'d

Issue #290, opened 2024-07-08
Enable equivalent  `iadmin modzonecollacl` functionality once it's fully understood


korydraughn:
- [ ] main
- [ ] 4-3-stable

PR #308, opened 2024-07-11
[#290] Remove API documentation for disabled zone operation

korydraughn:
The zone operation in question is disabled until we fully understand the iRODS API. Once we have a grasp on the API, we can revert this commit.


trel:
yep

Issue #272, opened 2024-04-08
Update GenQuery2 logic for iRODS 4.3.2


korydraughn:
- [x] /query endpoint
- [x] Docker-related files
- [x] Documentation / README, etc

## Build / Tools

iRODS 4.3.2 will ship with GenQuery2. The HTTP API needs to be able to switch between the different implementations of GenQuery2 due to the 4.2 compatibility option (i.e. 4.2 servers require the external GenQuery2 package).

korydraughn:
Because iRODS 4.3.2 ships with GenQuery2, the plan is to drop support for the GenQuery2 API plugin _(found at https://github.com/irods/irods_api_plugin_genquery2)_.

If users wish to continue using the GenQuery2 API plugin, they'll have to use HTTP API 0.3.0 (or earlier) and iRODS 4.3.1 (or earlier).

The reason for this is mostly due to:
- Conflicts between the two GenQuery2 implementations
- We have no plans for maintaining two implementations of GenQuery2

korydraughn:
Noting here that the `genquery2_enabled` property, returned by the `/info` endpoint, has been removed.

korydraughn:
Reopening. Forgot to update the test.

korydraughn:
Test added.

PR #319, opened 2024-08-03
[#272] Update dependencies to iRODS 4.3.2 for GenQuery2

korydraughn:

Docker-related changes work as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,24 +67,18 @@ API documentation can be found in [API.md](./API.md).
 
 ## Build Dependencies
 
-- iRODS development package
+- iRODS development package _(4.3.2 or later)_
 - iRODS externals package for boost
 - iRODS externals package for nlohmann-json
 - iRODS externals package for spdlog 
 - Curl development package
 - OpenSSL development package
-- GenQuery2 package (optional)
 
 ## Build
 
-This project can be built with or without support for [GenQuery2](https://github.com/irods/irods_api_plugin_genquery2).
+**IMPORTANT: As documented under [Build Dependencies](#build-dependencies), the minimum version requirement for the iRODS development package is 4.3.2 or later. As a result, support for the [GenQuery2 API plugin](https://github.com/irods/irods_api_plugin_genquery2) has been removed. To use GenQuery2, the iRODS HTTP API must be connected to a server running iRODS 4.3.2 or later.**
 
-GenQuery2 is disabled by default.
-
-### Building without GenQuery2
-
-If you don't need support for GenQuery2, follow the normal CMake steps.
-
+To build this project, follow the normal CMake build steps.
 ```bash
 mkdir build # Preferably outside of the repository
 cd build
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -124,22 +99,21 @@ The builder image is responsible for building the iRODS HTTP API package. Before
 docker build -t irods-http-api-builder -f irods_builder.Dockerfile .
 ```
 
-With the builder image in hand, all that's left is to get the source code for the GenQuery2 project and HTTP API project. The builder image is designed to compile code sitting on your machine. This is important because it gives you the ability to build any fork or branch of the projects.
+With the builder image in hand, all that's left is to build the iRODS HTTP API project. The builder image is designed to compile code sitting on your machine. This is important because it gives you the ability to build any version of the project derived from this branch.
```
alanking, line 102: I don't quite understand the last sentence:
> This is important because it gives you the ability to build any version of the project derived from this branch.

I think the advantage of being able to build from source code on the host filesystem is that you don't have to "sync" it anywhere (i.e. push to a remote repo in GitHub) in order to build it in the container(s). Being able to checkout a specific commit-ish to build can be done without the source code being on the host filesystem. Maybe I've misunderstood what this is saying.

endpoints/information/src/main.cpp
```diff
@@ -34,18 +34,11 @@ namespace irods::http::handler
...
-// NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
-#  define GENQUERY2_ENABLED false
-#endif // IRODS_ENABLE_GENQUERY2
-
 			// clang-format off
 			res.body() = json{
 				{"api_version", irods::http::version::api_version},
 				{"build", irods::http::version::sha},
-				{"genquery2_enabled", GENQUERY2_ENABLED},
+				{"genquery2_enabled", true},
```
alanking, line 41: Should we deprecate this field?

alanking (commented):
Looks good to me. Just had a couple thoughts on the README piece.

README.md
korydraughn: This is primarily about the removal of support for GenQuery2. If there are users who need GenQuery2 for iRODS 4.2.x, they can't use builder images derived from this commit.

alanking: Can we just say that instead? The sentence as written does not seem related to GenQuery2 or iRODS 4.2.x at all.

korydraughn: Sure. Will reword that.

endpoints/information/src/main.cpp
korydraughn: Whether we should deprecate or remove it isn't clear to me yet. I'm trying to think though whether there's value in keeping it.

Are users relying on it? What if GenQuery2 contains a nasty bug that forces us to disable it in the HTTP API? How do we communicate that? Should it be communicated through that property or just documentation?

The fact we haven't reached 1.0 status already provides signal that things can change a lot. Does that mean we can skip deprecating things? I'm pretty sure the interface has changed without notice in the past.

korydraughn: After some discussion, we're going to remove that property.

README.md
korydraughn: Adjusted wording on L102 and at the beginning of the `Build` section.

korydraughn: Will likely link the GenQuery2 API plugin repo so people know what we're referring to.

README.md
```diff
@@ -76,6 +76,8 @@ API documentation can be found in [API.md](./API.md).
 
 ## Build
 
+**IMPORTANT: As documented under [Build Dependencies](#build-dependencies), the minimum version requirement for the iRODS development package is 4.3.2 or later. As a result, support for the GenQuery2 API plugin has been removed. To use GenQuery2, the iRODS HTTP API must be connected to a server running iRODS 4.3.2 or later.**
```
alanking: What happens if the HTTP API is connected to an iRODS server running 4.3.1 or earlier? It falls back to GenQuery1?

alanking (commented):
Had one clarifying question but otherwise seems good

README.md
alanking: I was a little confused by the "To use GenQuery2" clause because my understanding is that the HTTP API is going to use GenQuery2 no matter what, but now it's going to use the one baked into the server which is only available in iRODS 4.3.2 and later, and therefore will not work with iRODS 4.3.1 and earlier.

If this is the case, I think this should say...

> The iRODS HTTP API must be connected to a server running iRODS 4.3.2 or later because the HTTP API now uses the GenQuery2 API which is built into the server rather than the GenQuery2 API plugin.

Instead of...

> To use GenQuery2, the iRODS HTTP API must be connected to a server running iRODS 4.3.2 or later.

Thoughts?

trel: I thought it knew which server version it was attached to... and behaved accordingly...?

korydraughn: Users can still run the HTTP API against 4.3.1. They just won't be able to invoke GenQuery2 _(even though the HTTP API has support for it built in due to it being compiled against 4.3.2 dev libs)_.

The HTTP API still defaults to GenQuery1 for the /query endpoint. If a user wants to use/call GenQuery2, they must set the appropriate parameter in the HTTP request and the HTTP API must be connected to a 4.3.2 (or later) server.

Only one implementation of GenQuery2 is supported because the GenQuery2 API plugin repo does not compile against 4.3.2. _Of course this could be made to work, but it didn't seem worth the effort at the end of the day._

alanking: Ah, okay, that makes sense. Yes, resolving.

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Pounded.

Issue #196, opened 2024-01-04
implement incremental write loop in `op_write` to avoid unresponsive server


korydraughn:
The following loop can lead to the server becoming unresponsive if enough clients execute long running write operations. Instead, the loop needs to post write operations to the background thread pool. This allows other requests to be served in a timely manner.
https://github.com/irods/irods_client_http_api/blob/f168903bc5dc946d47a43a6a307ff5fbc133f27c/endpoints/data_objects/src/main.cpp#L625-L637

PR #320, opened 2024-08-19
[#196] Implement incremental write loop for write operation

korydraughn:
All tests pass.

Considering the idea of passing an input struct rather than 10+ parameters. Not required, but much easier to deal with from a code standpoint and while reading. Can also handle this later.

TODO: Add tests which force the use of bg thread pool when writing to iRODS. The number of writes depend on the configuration of the HTTP API, so we'll need to document the assumptions of the test suite too.

### DIFF
--- a/endpoints/data_objects/src/main.cpp
+++ b/endpoints/data_objects/src/main.cpp
@@ -357,6 +357,124 @@ namespace
 		std::int64_t remaining_bytes_;
 	}; // incremental_read
 
+	class incremental_write : public std::enable_shared_from_this<incremental_write>
+	{
+	  public:
+		incremental_write(
+			irods::http::session_pointer_type& _sess_ptr,
+			unsigned int _http_version,
+			bool _http_keep_alive,
+			irods::http::connection_facade _conn,
+			std::unique_ptr<io::client::native_transport> _tp,
+			std::unique_ptr<io::odstream> _out,
+			io::odstream* _out_ptr,
+			std::unique_ptr<irods::at_scope_exit<std::function<void()>>> _mark_pw_stream_as_usable,
+			std::string _buffer,
+			std::int64_t _remaining_bytes,
+			std::int64_t _max_bytes_per_write,
+			bool _is_parallel_write)
+			: sess_ptr_{_sess_ptr->shared_from_this()}
+			, res_{http::status::ok, _http_version}
+			, conn_{std::move(_conn)}
+			, tp_{std::move(_tp)}
+			, out_{std::move(_out)}
+			, out_ptr_{_out_ptr}
+			, mark_pw_stream_as_usable_{std::move(_mark_pw_stream_as_usable)}
+			, buffer_{std::move(_buffer)}
+			, remaining_bytes_{_remaining_bytes}
+			, max_bytes_per_write_{_max_bytes_per_write}
+			, read_pos_{buffer_.data()}
+			, is_parallel_write_{_is_parallel_write}
+		{
+			res_.set(http::field::server, irods::http::version::server_name);
+			res_.set(http::field::content_type, "application/json");
+			res_.keep_alive(_http_keep_alive);
+		} // constructor
+
+		auto start() -> void
+		{
+			stream_bytes_to_irods();
+		} // start
+
+	  private:
+		auto stream_bytes_to_irods() -> void
+		{
+			irods::http::globals::background_task([self = shared_from_this(), fn = __func__]() mutable {
+				try {
+					if (self->remaining_bytes_ > 0) {
+						if (!*self->out_ptr_) {
+							logging::error(
+								*self->sess_ptr_,
+								"{}: Output stream is in a bad state. Client should restart the entire transfer.",
+								fn);
+							return self->sess_ptr_->send(
+								irods::http::fail(self->res_, http::status::internal_server_error));
+						}
+
+						const auto to_send =
+							std::min<std::streamsize>(self->remaining_bytes_, self->max_bytes_per_write_);
+						logging::debug(
+							*self->sess_ptr_,
+							"{}: Write buffer: remaining=[{}], sending=[{}].",
+							fn,
+							self->remaining_bytes_,
+							to_send);
+						self->out_ptr_->write(self->read_pos_, to_send);
+						self->read_pos_ += to_send; // NOLINT(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+						self->remaining_bytes_ -= to_send;
+
+						return self->stream_bytes_to_irods();
+					}
+
+					// If we're performing a normal write, close the stream before returning a response.
+					// This is required so that the iRODS server triggers appropriate policy before handing
+					// back control to the client. For example, replication resources and synchronous replication.
+					if (!self->is_parallel_write_) {
+						self->out_ptr_->close();
+					}
+
+					self->res_.body() = json{{"irods_response", {{"status_code", 0}}}}.dump();
+					self->res_.prepare_payload();
+					self->sess_ptr_->send(std::move(self->res_));
+				}
+				catch (const json::exception& e) {
+					logging::error(*self->sess_ptr_, "{}: {}", fn, e.what());
+					return self->sess_ptr_->send(irods::http::fail(self->res_, http::status::internal_server_error));
+				}
+				catch (const std::exception& e) {
+					logging::error(*self->sess_ptr_, "{}: {}", fn, e.what());
+					return self->sess_ptr_->send(irods::http::fail(self->res_, http::status::internal_server_error));
+				}
+			});
+		} // stream_bytes_to_irods
+
+		// The following member variables represent state initialized by op_write.
+		// Instances of this class require the state to last until after the final write
+		// operation completes or an error occurs, whichever happens first.
+		//
+		// Instances of this class own all state passed from op_write.
+
+		irods::http::session_pointer_type sess_ptr_;
+		http::response<http::string_body> res_;
+
+		irods::http::connection_facade conn_;
+		std::unique_ptr<io::client::native_transport> tp_;
+		std::unique_ptr<io::odstream> out_;
+		io::odstream* out_ptr_;
+
+		// A callable for signaling when a parallel-write stream is available for use.
+		std::unique_ptr<irods::at_scope_exit<std::function<void()>>> mark_pw_stream_as_usable_;
+
+		// The data to write to iRODS and information for tracking progress.
+		std::string buffer_;
+		std::int64_t remaining_bytes_;
+		std::int64_t max_bytes_per_write_;
+		const char* read_pos_;
+
+		// Indicates whether the client is performing a parallel write.
+		const bool is_parallel_write_;
+	}; // incremental_write
+
 	//
 	// Operation handler implementations
 	//
@@ -654,7 +772,7 @@ namespace
 		                                       client_info,
 		                                       _sess_ptr,
 		                                       _req = std::move(_req),
-		                                       _args = std::move(_args)] {
+		                                       _args = std::move(_args)]() mutable {
 			logging::info(*_sess_ptr, "{}: client_info.username = [{}]", fn, client_info.username);
 
 			http::response<http::string_body> res{http::status::ok, _req.version()};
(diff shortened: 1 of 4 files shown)


endpoints/data_objects/src/main.cpp
```diff
@@ -356,6 +356,100 @@ namespace
...
+		io::odstream* out_ptr_;
+
+		std::unique_ptr<irods::at_scope_exit<std::function<void()>>> mark_pw_stream_as_usable_;
+
+		std::string buffer_;
+		std::int64_t remaining_bytes_;
+		std::int64_t internal_buffer_size_;
+		const char* read_pos_;
+
+		const bool is_parallel_write_;
```
korydraughn, line 450: Document these

endpoints/data_objects/src/main.cpp
```diff
@@ -356,6 +356,100 @@ namespace
...
+			bool _http_keep_alive,
+			irods::http::connection_facade& _conn,
+			std::unique_ptr<io::client::native_transport>& _tp,
+			std::unique_ptr<io::odstream>& _out,
+			io::odstream* _out_ptr,
+			std::unique_ptr<irods::at_scope_exit<std::function<void()>>>& _mark_pw_stream_as_usable,
+			std::string& _buffer,
+			std::int64_t _remaining_bytes,
+			std::int64_t _internal_buffer_size,
+			bool _is_parallel_write)
```
korydraughn, line 374: Consider passing via an input struct.

korydraughn:
Added tests.

endpoints/data_objects/src/main.cpp
```diff
@@ -356,6 +356,100 @@ namespace
...
+
+				// If we're performing a normal write, close the stream before returning a response.
+				// This is required so that the iRODS server triggers appropriate policy before handing
+				// back control to the client. For example, replication resources and synchronous replication.
+				if (!self->is_parallel_write_) {
+					self->out_ptr_->close();
+				}
+
+				self->res_.body() =
+					json{{"irods_response", {{"status_code", 0}}}}.dump(); // TODO Could throw an exception.
```
korydraughn, line 428: Handle TODO

endpoints/data_objects/src/main.cpp
```diff
@@ -356,6 +356,100 @@ namespace
...
+					}
+
+					const auto to_send = std::min<std::streamsize>(self->remaining_bytes_, self->internal_buffer_size_);
+					logging::debug(
+						"{}: Write buffer: remaining=[{}], sending=[{}].", fn, self->remaining_bytes_, to_send);
+					self->out_ptr_->write(self->read_pos_, to_send);
+					self->read_pos_ += to_send;
+					self->remaining_bytes_ -= to_send;
+
+					return self->stream_bytes_to_irods();
```
alanking, line 417: So rather than having a single foreground task which loops to stream bytes, we've switched to a recursive background task which spawns background tasks from itself for each write (itself also being a background task) until there's nothing left to write. Is that right?

If so, why the switch to this model? Just trying to understand.

test/test_irods_http_api.py
```diff
@@ -1648,6 +1650,177 @@ def test_parallel_writes(self):
         self.assertEqual(r.status_code, 200)
         self.assertEqual(r.json()['irods_response']['status_code'], 0)
 
+    def test_parallel_writes_with_data_exceeding_internal_write_threshold(self):
+        # This test assumes the HTTP API is configured to use a value smaller than
+        # 96kb for "/irods_client/buffer_size_in_bytes_for_write_operations". This is
```
alanking, line 1655: If it's possible for the HTTP API to check its own configuration, we can skip or fail the test if it's not meeting the assumptions. Just a thought; don't have to do that.

endpoints/data_objects/src/main.cpp
```diff
@@ -356,6 +356,100 @@ namespace
...
+			, buffer_{std::move(_buffer)}
+			, remaining_bytes_{_remaining_bytes}
+			, internal_buffer_size_{_internal_buffer_size}
+			, read_pos_{buffer_.data()}
+			, is_parallel_write_{_is_parallel_write}
+		{
+			res_.set(http::field::server, irods::http::version::server_name);
+			res_.set(http::field::content_type, "application/json");
+			res_.keep_alive(_http_keep_alive);
+		}
```
alanking, line 391: Maybe add `// constructor` here?

test/test_irods_http_api.py
```diff
@@ -1648,6 +1650,177 @@ def test_parallel_writes(self):
...
+        data96kb = ('A' * 32768) + ('B' * 32768) + ('C' * 32768)
+
+        # Calculate a checksum for data.
+        # The checksum will be used to confirm the data was uploaded correctly.
+        checksum = base64.b64encode(hashlib.sha256(data96kb.encode('utf-8')).digest()).decode('utf-8')
+        self.logger.debug(f'checksum = [{checksum}]')
+
+        # Indicates whether the parallel_write_shutdown operation needs to be called.
+        # This is primarily used for clean up when errors occur.
+        invoke_parallel_write_shutdown = True
```
alanking, line 1681: Should we only set this to `True` after `parallel_write_init` has completed successfully? I think `parallel_write_handle` will not be defined should either of the assumptions following `parallel_write_init` fail, which would cause problems in the cleanup. Until `parallel_write_init` succeeds, I think it should be set to `False`.

Also, I would argue the word "primarily" can be removed from the comment as it doesn't appear to be used anywhere else besides the cleanup.

alanking (commented):
Seems good overall. Had a couple questions, couple suggestions

endpoints/data_objects/src/main.cpp
korydraughn: The switch was inevitable. If we don't break up long running tasks, then the server runs the risk of becoming unresponsive. There are a limited number of threads and they have to devote time to each request. If too many clients appear and each of them attempt to upload a file which causes a thread to have to enter a loop, clients will feel the latency.

test/test_irods_http_api.py
korydraughn: The tests do not have that ability right now. We can add that though.

That information could be exposed via the /info endpoint. The question then becomes, is that information valuable to users or is it only valuable to the tests.

Please open an issue for that.

endpoints/data_objects/src/main.cpp
korydraughn: Sure thing.

test/test_irods_http_api.py
alanking: Aren't these tests running on the same machine as the HTTP API server? If so, we could just crack open the config file, right? Sorry if these are naive questions.

korydraughn: It's fine. Ask all the questions.

> Aren't these tests running on the same machine as the HTTP API server?

They aren't required to run on the same machine. The config.py provides information for helping the tests run against local or remote HTTP API servers. The tests do not assume the HTTP API is local to them.

I run the tests locally, but others have been known to run/test it remotely.

alanking: Okay, that makes sense. Maybe we can set the configurations in question in the test config and leave a comment referencing this test should anybody ever change these? Just concerned about the fact that these tests can (and will?) pass even if the assumptions are not met.

Anyway, I created https://github.com/irods/irods_client_http_api/issues/322 for the configuration fetching mechanism.

korydraughn: That's a possibility. The GenQuery2 tests are controlled by config.py just like you've described.

endpoints/data_objects/src/main.cpp
korydraughn: We can revisit this at a later time. The current implementation works and it's unlikely that anyone will need to modify this chunk of code once merged.

I'll open an issue to revisit this at a later time.

test/test_irods_http_api.py
korydraughn: Resolving now that we have issue #322.

endpoints/data_objects/src/main.cpp
korydraughn: Done

test/test_irods_http_api.py
korydraughn: Done

endpoints/data_objects/src/main.cpp

endpoints/data_objects/src/main.cpp

endpoints/data_objects/src/main.cpp

endpoints/data_objects/src/main.cpp
korydraughn: Created #326.

korydraughn:
All review comments have been resolved.

Tests still pass.

Waiting for final review.

test/test_irods_http_api.py
```diff
@@ -1648,6 +1650,181 @@ def test_parallel_writes(self):
...
+            parallel_write_handle = result['parallel_write_handle']
+
+            # Make sure the parallel_write_shutdown operation is invoked now that we
+            # have a parallel-write handle.
+            invoke_parallel_write_shutdown = True
+
+            # Write to the data object using the parallel write handle.
+            futures = []
+            with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
+                chunk_size = len(data96kb) // 3
```
korydraughn, line 1704: 3 -> `stream_count`

korydraughn: Done

korydraughn:
Made minor tweak to test - no change in behavior.

Tests still pass.

Ready for final review again.

alanking (approved):
I'd feel better with one more set of eyeballs before #s. Up to you

trel:
looks good to me - i think ready for squashing.

trel (approved):
yep, confirmed it looks the same.   separation is good on the two commits.    

wait for one more before #ing

endpoints/data_objects/src/main.cpp
```diff
@@ -357,6 +357,124 @@ namespace
...
+		incremental_write(
+			irods::http::session_pointer_type& _sess_ptr,
+			unsigned int _http_version,
+			bool _http_keep_alive,
+			irods::http::connection_facade& _conn,
+			std::unique_ptr<io::client::native_transport>& _tp,
+			std::unique_ptr<io::odstream>& _out,
+			io::odstream* _out_ptr,
+			std::unique_ptr<irods::at_scope_exit<std::function<void()>>>& _mark_pw_stream_as_usable,
+			std::string& _buffer,
```
korydraughn, line 372: Could make it more obvious that ownership of some inputs is being transferred by requiring an R-value.

Making that change would only affect the call site. There would be no change in behavior.

endpoints/data_objects/src/main.cpp
```diff
@@ -826,57 +944,59 @@ namespace
 					return _sess_ptr->send(irods::http::fail(res, http::status::bad_request));
 				}
 
-				static const auto buffer_size =
+				static const auto max_number_of_bytes_per_write =
 					irods::http::globals::configuration()
 						.at(json::json_pointer{"/irods_client/buffer_size_in_bytes_for_write_operations"})
```
korydraughn, line 949: Consider renaming `/irods_client/buffer_size_in_bytes_for_write_operations` config option.

It is not a buffer size. It is the max number of bytes to write to iRODS in a single call.

endpoints/data_objects/src/main.cpp

endpoints/data_objects/src/main.cpp
korydraughn: Renamed to `max_number_of_bytes_per_write_operation`.

Uses the same naming scheme as the option for read operations.

korydraughn:
Tests still pass.

The latest commits do not introduce any changes in behavior.

trel:
good new name.   looks good.

alanking (commented):
Squash em, I say

korydraughn:
Squashed.

@MartinFlores751's review is on the way.

MartinFlores751 (approved):
Looks good to me.

To note, Kory and I did talk yesterday about the 'boost idiom' of the 'recursion with a shared pointer to self' just to understand why it was used vs putting something similar to the previous code in a background task instead. Similar question: https://github.com/irods/irods_client_http_api/pull/320#discussion_r1725022659

alanking (approved):
\# it if happy with the changes.

Issue #323, opened 2024-08-21
zones/stat is not in API.md


trel:
## Bug Report

missing zones/stat operation in API.md documentation

PR #327, opened 2024-08-21
[#323] Document stat operation for /zones endpoint

korydraughn:


trel:
seems good - easy to tweak if we missed something.

korydraughn:
Pounded.

Issue #343, opened 2024-10-03
Provide local validation of OAuth 2.0 JWT Access Tokens


MartinFlores751:
Using the OAuth 2.0 Access Token JWT Profile[^1], we should be able to validate access tokens for OpenID providers who give JWT access tokens but don't provide an introspection endpoint. This should cover most OpenID providers, though some providers may not provide either standard methods.

[^1]: https://www.rfc-editor.org/rfc/rfc9068

korydraughn:
Does this overlap with any existing OAuth issues?

MartinFlores751:
Not from what I can tell. It's a bit adjacent to the validating tokens issue, though that's focused on our 'client mode'. This issue is focused on 'protected resource mode'.

korydraughn:
Okay. Placing in the 0.5.0 milestone for now.

Can bump it if necessary.

PR #344, opened 2024-10-21
[#343] Validate JWT Access Tokens locally

MartinFlores751:
This PR provides the ability to validate JWT Access Tokens locally. This should allow for greater support of various OpenID Providers.


endpoints/authentication/src/main.cpp
MartinFlores751, line 1: Remove these after initial testing.

core/src/common.cpp
```diff
@@ -407,14 +667,27 @@ namespace irods::http
...
-				auto json_res{hit_introspection_endpoint(url_encode_body(args))};
+					if (!possible_json_res) {
+						return {.response = fail(status_type::unauthorized)};
+					}
 
-				// Validate access token
-				if (!json_res.at("active").get<bool>()) {
-					logging::warn("{}: Access token is invalid or expired.", __func__);
-					return {.response = fail(status_type::unauthorized)};
+					json_res = *possible_json_res;
```
MartinFlores751, line 690: We probably want to default to local validation. Either that or make order configurable?

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+
+		// Begin building up the JWT verifier...
+		auto verifier{
+			jwt::verify<jwt::traits::nlohmann_json>()
+			// Token MUST have issuer match what is defined by the OpenID Provider
+				.with_issuer(
+					irods::http::globals::oidc_endpoint_configuration().at("issuer").get_ref<const std::string&>())
+			// 'aud' MUST contain identifier we expect (ourselves)
+				.with_audience(
+					irods::http::globals::oidc_configuration().at("client_id").get_ref<const std::string&>())};
```
MartinFlores751, line 590: Don't to forget to add in the optional time leeway.

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+	/// See RFC 9068 for more details on JWT for OAuth 2.0
+	/// See RFC 7515 for details on JSON Web Signature (JWS)
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in] _jwt A std::string represeting the JWT to verify.
+	///
+	/// \returns The JWT payload if the token can be validated. Otherwise, an empty std::optional is returned
+	///
+	/// \todo Handle encrypted tokens. MUST reject unencrypted tokens if encryption was negotiated at registration.
+	/// \date 2024-10-16 jwt-cpp does not support encrypted tokens as of current.
```
MartinFlores751, line 557: Do we like this sort of comment in doxygen?

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+		auto decoded_token{jwt::decode<jwt::traits::nlohmann_json>(_jwt)};
+
+		// Parse the JWKs discoverd from the OpenID Provider
+		static auto jwks{jwt::parse_jwks<jwt::traits::nlohmann_json>(fetch_jwks_from_openid_provider())};
+
+		// Manually verify 'typ' matches what is specified in RFC 9068
+		// Validate 'typ' is 'at+jwt' or 'application/at+jwt', reject all others
+		try {
+			auto token_type{decoded_token.get_header_claim("typ").as_string()};
+			if (!(token_type == "at+jwk" || token_type == "application/at+jwk")) {
```
MartinFlores751, line 571: Double check that this is case sensitive.

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+			auto y{_jwk.get_jwk_claim("y").as_string()};
+
+			// Add verification algorithm
+			_verifier.allow_algorithm(jwt::algorithm::es256(jwt::helper::create_public_key_from_ec_components(crv, x, y)));
+		}
+	}
+
+	/// Bleh blah
+	///
+	/// \returns A refernce to the provided jwt::verifier, allowing for chaining.
```
MartinFlores751, line 493: If you return a reference, consider adding a comment about object lifetime?

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+		}
+		catch (const std::runtime_error& e) {
+			// We cannot pick out the specific key used...
+			// Find JWKs that match any of the following:
+			//   - 'alg' (optional) matches exactly
+			//   - 'kty' (required) matches 'alg' family [EC,RSA,oct]
+			//   - 'use' (required if sign & encrypt keys exist) to be 'sig'
+			//
+			// Additionally, symetric algoritms it /seems/ that
+			// "... the octects of the 'client_secret'..." are used for the algorithm...
```
MartinFlores751, line 511: Test and verify how exactly the symmetric algorithms work out. Is there a `k` provided or not?

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+				search_string = "RSA";
+			}
+			// EC family
+			else if (algorithm_family == "ES") {
+				search_string = "EC";
+			}
+			// Symmetric algo
+			else if (algorithm_family == "HS") {
+				search_string = "oct";
+			}
```
MartinFlores751, line 526: Comment where you got these 'magic values' from.

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+
+			// Get x coordinate parameter (JWA)
+			auto x{_jwk.get_jwk_claim("x").as_string()};
+
+			// Get y coordinate parameter (JWA)
+			auto y{_jwk.get_jwk_claim("y").as_string()};
+
+			// Add verification algorithm
+			_verifier.allow_algorithm(jwt::algorithm::es256(jwt::helper::create_public_key_from_ec_components(crv, x, y)));
+		}
```
MartinFlores751, line 488: Given the params for a family of algorithms are the same, get the family, extract the params, then figure out the algorithm?

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+		req.set(beast::http::field::user_agent, irods::http::version::server_name);
+		req.set(beast::http::field::accept, "application/json");
+		req.prepare_payload();
+
+		// Send request & receive response
+		auto res{tcp_stream->communicate(req)};
+
+		logging::debug("{}: Received the following response: [{}]", __func__, res.body());
+
+		// JSONize response
```
MartinFlores751, line 435: Remove old comment.

korydraughn:
Good stuff. Will review when it's out of draft or feedback is requested.

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+
+		const auto parsed_uri{boost::urls::parse_uri(jwks_uri)};
+
+		if (parsed_uri.has_error()) {
+			logging::error(
+				"{}: Error trying to parse jwks_uri [{}]. Please check configuration.",
+				__func__,
+				jwks_uri);
+			return "{\"error\", \"bad endpoint\"}";
+		}
```
MartinFlores751, line 412: Figure out error handling for this.

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+		return _verifier;
+	}
+	
+	/// Validates an OAuth 2.0 Access Token using
+	///
+	/// See RFC 9068 for more details on JWT for OAuth 2.0
+	/// See RFC 7515 for details on JSON Web Signature (JWS)
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in] _jwt A std::string represeting the JWT to verify.
```
trel, line 552:
```suggestion
	/// \param[in] _jwt A std::string representing the JWT to verify.
```

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+					return;
+				}
+			});
+		}
+
+		// Allow for chaining
+		return _verifier;
+	}
+	
+	/// Validates an OAuth 2.0 Access Token using
```
trel, line 546: need an end to this sentence

core/include/irods/private/http_api/common.hpp
```diff
@@ -169,6 +169,8 @@ namespace irods::http
 
 	auto map_json_to_user(const nlohmann::json& _json) -> std::optional<std::string>;
 
+	auto validate_using_local_validation(std::string _thing) -> std::optional<nlohmann::json>;
```
trel, line 172: perhaps... `_token` or `_jwt` instead of `_thing`?

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+	///
+	/// See RFC 9068 for more details on JWT for OAuth 2.0
+	/// See RFC 7515 for details on JSON Web Signature (JWS)
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
```
MartinFlores751, line 550: Consider using RFC XXXX [DESCRIPTION] ([SHORT HAND]), and then reference in function using [SHORT HAND].

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+		try {
+			auto token_type{decoded_token.get_header_claim("typ").as_string()};
+			if (!(token_type == "at+jwk" || token_type == "application/at+jwk")) {
+				logging::error("{}: Access Token with [typ] of type [{}] is not supported.", __func__, token_type);
+				// return std::nullopt;
+			}
+		}
+		// Handling missing 'typ'
+		catch (const jwt::error::claim_not_present_exception& e) {
+			logging::error("{}: invalid JWT [{}]", __func__, e.what());
```
MartinFlores751, line 578: Be more specific, 'invalid access token' or use the full `typ` name.

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+		// Send request & receive response
+		auto res{tcp_stream->communicate(req)};
+
+		logging::debug("{}: Received the following response: [{}]", __func__, res.body());
+
+		// JSONize response
+		return res.body();
+	}
+
+	/// Adds the specified algorithm \p _alg to the verifier \p _verifier, using the additional informaiton provided by the JWK \p _jwk.
```
trel, line 439:
```suggestion
	/// Adds the specified algorithm \p _alg to the verifier \p _verifier, using the additional information provided by the JWK \p _jwk.
```

core/src/common.cpp
MartinFlores751, line 571: Comment does not match up with code, figure out which is real.

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+	///
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in,out] _verifier The jwt::verifier to add additional verification algorithms to.
+	/// \param[in]     _jwk      The JWK containing the required JWA information.
+	/// \param[in]     _alg      The sigining algorithm requested by the signed JWT.
```
trel, line 445:
```suggestion
	/// \param[in]     _alg      The signing algorithm requested by the signed JWT.
```

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+	///
+	/// \todo Should this be separated out by algorithm familiy?
```
trel, line 447:
```suggestion
	/// \todo Should this be separated out by algorithm family?
```

MartinFlores751 (commented):
'Off the top of my head' comments.

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+			// Get modulus parameter (JWA)
+			auto mod{_jwk.get_jwk_claim("n").as_string()};
+
+			// Get exponent parameter (JWA)
+			auto exp{_jwk.get_jwk_claim("e").as_string()};
+
+			// Add verification algorithm
+			_verifier.allow_algorithm(jwt::algorithm::rs256(jwt::helper::create_public_key_from_rsa_components(mod, exp)));
+		}
+		// For symetric excryption, no parameter 'k' is provided
```
trel, line 463:
```suggestion
		// For symmetric encryption, no parameter 'k' is provided
```

core/src/common.cpp
trel, line 488: and if the algorithm doesn't match...   it falls through without doing anything?

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+
+			// Get y coordinate parameter (JWA)
+			auto y{_jwk.get_jwk_claim("y").as_string()};
+
+			// Add verification algorithm
+			_verifier.allow_algorithm(jwt::algorithm::es256(jwt::helper::create_public_key_from_ec_components(crv, x, y)));
+		}
+	}
+
+	/// Bleh blah
```
trel, line 491: more!

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+			add_alg_from_jwk(_verifier, jwk, _alg);
+		}
+		catch (const std::runtime_error& e) {
+			// We cannot pick out the specific key used...
+			// Find JWKs that match any of the following:
+			//   - 'alg' (optional) matches exactly
+			//   - 'kty' (required) matches 'alg' family [EC,RSA,oct]
+			//   - 'use' (required if sign & encrypt keys exist) to be 'sig'
+			//
+			// Additionally, symetric algoritms it /seems/ that
```
trel, line 510:
```suggestion
			// Additionally, symmetric algorithms it /seems/ that
```

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+			// "... the octects of the 'client_secret'..." are used for the algorithm...
+			const auto algorithm_family{_alg.substr(0,2)};
+			std::string search_string;
+
+			// RSA family
+			if (algorithm_family == "RS") {
+				search_string = "RSA";
+			}
+			// EC family
+			else if (algorithm_family == "ES") {
```
trel, line 520: ES is correct?

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+	///
+	/// \todo Handle encrypted tokens. MUST reject unencrypted tokens if encryption was negotiated at registration.
+	/// \date 2024-10-16 jwt-cpp does not support encrypted tokens as of current.
+	auto validate_using_local_validation(std::string _jwt) -> std::optional<nlohmann::json> {
+		namespace logging = irods::http::log;
+
+		// Decode the JWT
+		auto decoded_token{jwt::decode<jwt::traits::nlohmann_json>(_jwt)};
+
+		// Parse the JWKs discoverd from the OpenID Provider
```
trel, line 564:
```suggestion
		// Parse the JWKs discovered from the OpenID Provider
```

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+		static auto jwks{jwt::parse_jwks<jwt::traits::nlohmann_json>(fetch_jwks_from_openid_provider())};
+
+		// Manually verify 'typ' matches what is specified in RFC 9068
+		// Validate 'typ' is 'at+jwt' or 'application/at+jwt', reject all others
+		try {
+			auto token_type{decoded_token.get_header_claim("typ").as_string()};
+			if (!(token_type == "at+jwk" || token_type == "application/at+jwk")) {
+				logging::error("{}: Access Token with [typ] of type [{}] is not supported.", __func__, token_type);
+				// return std::nullopt;
```
trel, line 573: please confirm or delete

core/src/common.cpp
```diff
@@ -362,6 +364,264 @@ namespace irods::http
...
+			jwt::verify<jwt::traits::nlohmann_json>()
+			// Token MUST have issuer match what is defined by the OpenID Provider
+				.with_issuer(
+					irods::http::globals::oidc_endpoint_configuration().at("issuer").get_ref<const std::string&>())
+			// 'aud' MUST contain identifier we expect (ourselves)
+				.with_audience(
+					irods::http::globals::oidc_configuration().at("client_id").get_ref<const std::string&>())};
+
+		// TODO: Refactor into separate function, taking verifier, jwk, token(?) as parameter, returning verifier(?)
+		// Details of the specific algoritms are defined by RFC 7518 (JWA)
```
trel, line 593:
```suggestion
		// Details of the specific algorithms are defined by RFC 7518 (JWA)
```

core/src/common.cpp
korydraughn: I'm assuming you're asking about the `\todo`. If so, then no.

If the TODO needs to be handled at a later time, open an issue and link it to this comment. That way you'll know the origin of the issue along with the highlighted code block. Include enough information in the issue so that you or another dev is able to understand the situation.

core/src/common.cpp
MartinFlores751: We should just log that the algorithm is not supported, and continue.

core/src/common.cpp
MartinFlores751: Yep. Going to push up commit with more comments soon.

core/src/common.cpp
MartinFlores751: How about the comment on the following line:

> 	/// \date 2024-10-16 jwt-cpp does not support encrypted tokens as of current.

core/src/common.cpp
MartinFlores751: Case-insensitive.

core/src/common.cpp
MartinFlores751: Commented for debugging. Uncommenting for later.

core/src/common.cpp
MartinFlores751: Did a `to_lower_copy(...)` before.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
korydraughn: What's the motivation for including that info?

MartinFlores751: Mostly to not forget the reason for omitting this feature when we eventually get back to it.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
korydraughn: Let's put it in an issue if it's something we'll need to address later. No need to have it in the code.

Link it here so we know origin of the issue.

Forgot to mention that if having a TODO in the code is needed, that's fine. You just need to apply the rules defined at https://docs.irods.org/4.3.3/developers/contributing/#todo-comments-in-irods-code.

Let's avoid the doxygen `\todo` directive.

core/src/common.cpp
```diff
@@ -520,7 +569,7 @@ namespace irods::http
 
 		// The first two characters of '_alg' should give us
```
trel, line 570:
```suggestion
		// The first two characters of 'alg' should give us
```

core/src/common.cpp
```diff
@@ -598,71 +655,98 @@ namespace irods::http
...
 
-		// Decode the JWT
-		auto decoded_token{jwt::decode<jwt::traits::nlohmann_json>(_jwt)};
+		try {
+			// Decode the JWT
+			auto decoded_token{jwt::decode<jwt::traits::nlohmann_json>(_jwt)};
 
-		// Parse the JWKs discoverd from the OpenID Provider
-		static auto jwks{jwt::parse_jwks<jwt::traits::nlohmann_json>(fetch_jwks_from_openid_provider())};
+			// Parse the JWKs discoverd from the OpenID Provider
```
trel, line 662:
```suggestion
			// Parse the JWKs discovered from the OpenID Provider
```

core/src/common.cpp
```diff
@@ -700,20 +784,18 @@ namespace irods::http
...
 			}
 
 			// If we're running as a protected resource, assume we have a OIDC token
-			if (oidc_config_iter->at("mode").get_ref<const std::string&>() == "protected_resource") {
+			if (irods::http::globals::oidc_configuration().at("mode").get_ref<const std::string&>() == "protected_resource") {
 				nlohmann::json json_res;
 
 				// Use introspection endpoint if it exsists
-				if (auto introspection_endpoint_iter{irods::http::globals::oidc_endpoint_configuration().find(nlohmann::json::json_pointer{"/introspection_endpoint"})}; introspection_endpoint_iter != std::end(irods::http::globals::oidc_endpoint_configuration())) {
+				if (auto introspection_endpoint_iter{irods::http::globals::oidc_endpoint_configuration().find(nlohmann::json::json_pointer{"/introspection_endpoint"})}; introspection_endpoint_iter != std::end(irods::http::globals::oidc_endpoint_configuration()) && false) {
```
trel, line 798: something weird going on with this line - has a semicolon.... and then i feel like we need a comment / explanation about that `&& false` at the end.

core/src/common.cpp
```diff
@@ -700,20 +784,18 @@ namespace irods::http
...
 				logging::error("{}: Could not find bearer token matching [{}].", __func__, bearer_token);
 				return {.response = fail(status_type::unauthorized)};
 			}
 
 			// If we're running as a protected resource, assume we have a OIDC token
-			if (oidc_config_iter->at("mode").get_ref<const std::string&>() == "protected_resource") {
+			if (irods::http::globals::oidc_configuration().at("mode").get_ref<const std::string&>() == "protected_resource") {
 				nlohmann::json json_res;
 
 				// Use introspection endpoint if it exsists
```
trel, line 797:
```suggestion
				// Use introspection endpoint if it exists
```

core/src/common.cpp
MartinFlores751: `&& false` is a debugging artifact. Was a bit lazy on flipping the execution order (i.e., try to validate locally, fall back to introspection endpoint).

core/src/common.cpp
MartinFlores751: `k` is not provided. The algorithm will likely not appear in the JWKs. Key may need to be provided separately.

core/src/common.cpp
MartinFlores751: Done.

endpoints/authentication/src/main.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Changed to execute both validation methods (if possible), starting with local validation and failing to the introspection endpoint if it is available.

The current implementation means both methods are exhausted, even if a token is invalid (e.g., expired, garbage).

core/src/common.cpp
```diff
@@ -362,6 +364,439 @@ namespace irods::http
...
+	}
+
+	/// Fetches JWKs from the location specified by the OpenID Provider
+	///
+	/// See OpenID Connect Discovery 1.0 Section 3 for info on jwks_uri
+	/// See RFC 7517 for more information on JSON Web Key (JWK)
+	///
+	/// \returns A std::string represting the JWKs from the OpenID Provider
+	///
+	/// \todo Stash the results from this function, only allow to run once.
```
MartinFlores751, line 396: Remove TODO, static acts as stash.

At most, consider moving to OIDC init. This will grow a new global, and it is unlikely the `jwks_uri` is improperly configured.

Additionally, having it as it is now could allow for retry behavior?

core/src/common.cpp
```diff
@@ -362,6 +364,439 @@ namespace irods::http
...
+
+		const auto jwks_uri{
+			irods::http::globals::oidc_endpoint_configuration().at("jwks_uri").get_ref<const std::string&>()};
+
+		const auto parsed_uri{boost::urls::parse_uri(jwks_uri)};
+
+		if (parsed_uri.has_error()) {
+			logging::error("{}: Error trying to parse jwks_uri [{}]. Please check configuration.", __func__, jwks_uri);
+			return "{\"error\", \"bad endpoint\"}";
+		}
```
MartinFlores751, line 409: Consider throwing here instead.

If we return, the error will be stashed in a static variable...

core/src/common.cpp
```diff
@@ -362,6 +364,439 @@ namespace irods::http
...
+		if (auto realm_secret{irods::http::globals::oidc_configuration().find("realm_secret")};
+		    realm_secret != std::end(irods::http::globals::oidc_configuration()))
+		{
+			key = realm_secret->get<std::string>();
+		}
+		else if (auto secret{irods::http::globals::oidc_configuration().find("client_secret")};
+		         secret != std::end(irods::http::globals::oidc_configuration()))
+		{
+			key = secret->get<std::string>();
+		}
```
MartinFlores751, line 463: Add comments explaining reasoning behind `realm_secret` and `client_secret` being used.

core/src/common.cpp
```diff
@@ -362,6 +364,439 @@ namespace irods::http
...
+
+	/// Adds the specified asymmetric algorithm \p _alg to the verifier \p _verifier, using the additional information
+	/// provided by the JWK \p _jwk.
+	///
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in,out] _verifier The jwt::verifier to add additional verification algorithms to.
+	/// \param[in]     _jwk      The JWK containing the required JWA information.
+	/// \param[in]     _alg      The signing algorithm requested by the signed JWT.
+	auto add_asymmetric_alg_from_jwk(
```
MartinFlores751, line 492: Currently, if `add_asymmetric_alg_from_jwk(...)` is passed an invalid JWK such that `.get_jwk_claim(...)` fails, the exception will bubble up and prevent any further action being done on the JWT to be validated. Is this desired?

I think so, but I'll leave this here to think about it later.

core/src/common.cpp
```diff
@@ -362,6 +364,439 @@ namespace irods::http
...
+			else if (algorithm_specifics == "384") {
+				logging::trace("{}: Adding [{}] to allowed verification algorithms.", __func__, _alg);
+				_verifier.allow_algorithm(jwt::algorithm::hs384(key));
+			}
+			else if (algorithm_specifics == "512") {
+				logging::trace("{}: Adding [{}] to allowed verification algorithms.", __func__, _alg);
+				_verifier.allow_algorithm(jwt::algorithm::hs512(key));
+			}
+			return;
+		}
```
MartinFlores751, line 479: Symmetric algorithms still need to be tested.

MartinFlores751 (commented):
Some comments and notes.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/include/irods/private/http_api/common.hpp
MartinFlores751: Removed, not used outside of `common.cpp`.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Removed.

core/src/common.cpp
MartinFlores751: Removed.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Removed.

core/src/common.cpp
MartinFlores751: Comment now describes algorithm types, where it gets shorthand names from.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Uncommented.

core/src/common.cpp
MartinFlores751: Removed.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Removed & reordered.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Tested & works.

core/src/common.cpp
MartinFlores751: See: https://github.com/irods/irods_client_http_api/pull/344#discussion_r1821737291

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Removed comment. We'll see if this is a desired feature later on.

core/src/common.cpp
```diff
@@ -362,6 +364,434 @@ namespace irods::http
...
+		auto algorithm_specifics{_alg.substr(2)};
+
+		// Extract the secret key
+		// OpenID secret should be our secret
+		std::string key;
+		if (auto realm_secret{irods::http::globals::oidc_configuration().find("realm_secret")};
+		    realm_secret != std::end(irods::http::globals::oidc_configuration()))
+		{
+			key = jwt::base::decode<jwt::alphabet::base64url>(jwt::base::pad<jwt::alphabet::base64url>(realm_secret->get<std::string>()));
+		}
```
MartinFlores751, line 456: Should we have `realm_secret` be a base64url or base64 encoded string?

Defaulted to base64url as that's what keycloak uses. The key should be encoded, as not all keys are ASCII printable.

core/src/common.cpp
MartinFlores751: Let's skip this and add later if desired.

core/src/common.cpp
MartinFlores751: No need to comment, interface should prevent rvalue being directly passed in.

MartinFlores751 (commented):
One last comment.

core/src/common.cpp
MartinFlores751, line 1: Is it worth moving the OAuth/OpenID functions out of `common.cpp`?

core/src/common.cpp
korydraughn: > Is this desired?

What are the pros and cons of letting it bubble up?
What does the admin see?
What does the client see?
What should the client do if they provide an invalid JWK?

core/src/common.cpp
korydraughn: > Should we have `realm_secret` be a base64url or base64 encoded string?

Is base64 encoding it a requirement?
What are the pros/cons of each encoding?

What is the `realm_secret` for?

> Defaulted to base64url as that's what keycloak uses. The key should be encoded, as not all keys are ASCII printable.

Is the decision to base64 encode the key strictly for printing it?

core/src/common.cpp
korydraughn: Yes. Just makes it easier for people to know which functions are specific to OIDC.

How many functions exist in common.cpp for OIDC?

core/src/common.cpp
trel: being able to print it helps with debugging... 

are there other use cases?

trel:
please clang-format and clang-tidy in two commits - don't squash anything yet please.

core/src/common.cpp
MartinFlores751: Why we need an encoding of some sort:

``` python
>>> # This a base64url encoded key from keycloak
>>> base64.urlsafe_b64decode('ACu4YsAjVil7KC3G9dVL_THazeFa4AVDZ83lNPiKV4KxEXc9tzrfeUzGJVvGsNG2oWuSLq18tIo4i8VUH6Dkpw==')
b'\x00+\xb8b\xc0#V){(-\xc6\xf5\xd5K\xfd1\xda\xcd\xe1Z\xe0\x05Cg\xcd\xe54\xf8\x8aW\x82\xb1\x11w=\xb7:\xdfyL\xc6%[\xc6\xb0\xd1\xb6\xa1k\x92.\xad|\xb4\x8a8\x8b\xc5T\x1f\xa0\xe4\xa7'
>>> # The key is not 'ASCII Printable'.
>>> base64.urlsafe_b64decode('ACu4YsAjVil7KC3G9dVL_THazeFa4AVDZ83lNPiKV4KxEXc9tzrfeUzGJVvGsNG2oWuSLq18tIo4i8VUH6Dkpw==').decode('ascii')
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xb8 in position 2: ordinal not in range(128)
```

`realm_secret` needs to be given to us through the config. I'm not sure the JSON parser can deal with the gibberish output. Additionally, it may be prone to errors if the key were to change at some point.

core/src/common.cpp
```diff
@@ -362,6 +364,434 @@ namespace irods::http
...
+		auto algorithm_family{_alg.substr(0, 2)};
+		auto algorithm_specifics{_alg.substr(2)};
+
+		// Extract the secret key
+		// OpenID secret should be our secret
+		std::string key;
+		if (auto realm_secret{irods::http::globals::oidc_configuration().find("realm_secret")};
+		    realm_secret != std::end(irods::http::globals::oidc_configuration()))
+		{
+			key = jwt::base::decode<jwt::alphabet::base64url>(jwt::base::pad<jwt::alphabet::base64url>(realm_secret->get<std::string>()));
```
alanking: Can we use `get_ref` here? Looks like it, but, I'm not familiar with the library being used here. :)

core/src/common.cpp
```diff
@@ -362,6 +364,434 @@ namespace irods::http
...
+			}
+			else if (algorithm_specifics == "384") {
+				logging::trace("{}: Adding [{}] to allowed verification algorithms.", __func__, _alg);
+				_verifier.allow_algorithm(jwt::algorithm::hs384(key));
+			}
+			else if (algorithm_specifics == "512") {
+				logging::trace("{}: Adding [{}] to allowed verification algorithms.", __func__, _alg);
+				_verifier.allow_algorithm(jwt::algorithm::hs512(key));
+			}
+			return;
```
alanking, line 477: What happens in the `else` case? It is neither added to the allowed algorithms, nor is a warning printed. It feels like it should drop through here or log a warning, rather than just `return`.

core/src/common.cpp
```diff
@@ -362,6 +364,434 @@ namespace irods::http
...
+				else if (algorithm_specifics == "384") {
+					logging::trace("{}: Adding [{}] to allowed verification algorithms.", __func__, _alg);
+					_verifier.allow_algorithm(jwt::algorithm::ps384(pub_key));
+				}
+				else if (algorithm_specifics == "512") {
+					logging::trace("{}: Adding [{}] to allowed verification algorithms.", __func__, _alg);
+					_verifier.allow_algorithm(jwt::algorithm::ps512(pub_key));
+				}
+			}
+			return;
```
alanking, line 544: Same thing here as above. The `else` cases above are unhandled and this will not fall through to the log statement below.

core/src/common.cpp
```diff
@@ -362,6 +364,434 @@ namespace irods::http
...
+	/// Adds verification algorithm(s) to the \p _verifier based on either the given 'kid' or algorthims matching
+	/// the family of the desired algorithm \p _alg.
+	///
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in,out] _verifier The jwt::verifier to add additional verification algorithms to.
+	/// \param[in]     _jwks     The JWKs to search through.
+	/// \param[in]     _jwt      The decoded JWT that needs to be verified.
+	///
+	/// \returns A refernce to the provided jwt::verifier, \p _verifier, allowing for chaining.
```
alanking, line 590:
```suggestion
	/// \returns A reference to the provided jwt::verifier, \p _verifier, allowing for chaining.
```

core/src/common.cpp
```diff
@@ -362,6 +364,434 @@ namespace irods::http
...
+	auto add_algorithms_to_verifier(
+		jwt::verifier<jwt::default_clock, jwt::traits::nlohmann_json>& _verifier,
```
alanking, line 592: Would a type alias be appropriate for this since it has popped up in many places?

core/src/common.cpp
```diff
@@ -362,6 +364,435 @@ namespace irods::http
...
+					}
+
+					logging::trace("{}: JWK [alg] does not match JWT [alg], ignoring.", __func__);
+					return;
+				}
+
+				// Fallback to required claim
+				if (_jwk.has_key_type()) {
+					// Extract the 'kty' of the JWK, compare to desired 'kty'
+					if (_jwk.get_key_type() == search_string) {
```
alanking, line 671: It is possible for `search_string` to be empty at this point. I guess it's not really a problem, just wondering if there's any short-circuiting opportunity. :)

core/src/common.cpp
```diff
@@ -362,6 +364,435 @@ namespace irods::http
...
+	///
+	/// See RFC 9068 for more details on JWT for OAuth 2.0 (OJWT)
+	/// See RFC 7516 for details on JSON Web Encryption (JWE)
+	/// See RFC 7515 for details on JSON Web Signature (JWS)
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in] _jwt A jwt::decoded_jwt<jwt::traits::nlohmann_json> representing the JWT to verify.
+	///
+	/// \returns The JWT payload if the token can be validated. Otherwise, an empty std::optional is returned
+	auto validate_using_local_validation(jwt::decoded_jwt<jwt::traits::nlohmann_json>& _jwt)
```
alanking, line 698: Should the input parameter be `const`?

alanking (commented):
Very nice work. The comments referencing the RFC sections relevant to that section of code are fantastic

core/src/common.cpp
korydraughn: Got it. Carry on.

core/src/common.cpp
```diff
@@ -397,23 +827,43 @@ namespace irods::http
 			// It's possible that the admin didn't include the OIDC configuration stanza.
 			// This use-case is allowed, therefore we check for the OIDC configuration before
 			// attempting to access it. Without this logic, the server would crash.
-			const auto oidc_config_iter =
-				config.find(nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"});
-			if (oidc_config_iter == std::end(config)) {
+			if (static const auto oidc_conf_exists{config.contains(nlohmann::json::json_pointer{"/http_server/authentication/openid_connect"})}; !oidc_conf_exists) {
```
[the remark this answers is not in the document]
korydraughn: Agreed. Let's keep static variables outside the if-stmts.

core/src/common.cpp
MartinFlores751: Alright, we'll leave as a base64url encoded string, and leave a comment on the readme.

core/src/common.cpp
MartinFlores751: Around eight, including changes in this PR.

core/src/common.cpp
korydraughn: Yep, please resolve once the comment is added to the README.

core/src/common.cpp
MartinFlores751: Failing to match something should be an error higher up. Good catch!

core/src/common.cpp
MartinFlores751: I'm terrible at names. `nlohmann_json_verifier`? Just `jwt_verifier`?

Suggestions appreciated.

alanking: I personally think `jwt_verifier` is fine unless we plan on supporting / implementing other kinds of `jwt::verifier` types in the future. What you have now is also just fine.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Fixed.

core/src/common.cpp
```diff
@@ -397,23 +827,43 @@ namespace irods::http
...
+					auto possible_json_res{validate_using_local_validation(token)};
+
+					if (possible_json_res) {
+						json_res = *possible_json_res;
+					}
+				}
+				// Parsing of the token failed, this is not a JWT access token
+				catch (const std::exception& e) {
+					logging::info("{}: {}", __func__, e.what());
+				}
```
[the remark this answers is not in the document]
MartinFlores751: Would lowering it further to trace be fine? Since this is an artifact of design, I feel like it's likely to be a low priority log.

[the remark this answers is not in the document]
trel: trace is usually only for letting a reader know the flow... no values shown/logged.

`debug` here feels... plenty?

core/src/common.cpp
```diff
@@ -362,6 +364,434 @@ namespace irods::http
 		return std::nullopt;
 	}
 
+	/// Validates an OAuth 2.0 Access Token using the Introspection Endpoint
+	/// See RFC 7662 on OAuth 2.0 Token Introspection for more details
+	///
+	/// \returns An optional containing a nlohmann::json object if verification was successful. Otherwise,
+	///          an empty optional is returned.
+	auto validate_using_introspection_endpoint(std::string _bearer_token) -> std::optional<nlohmann::json>
```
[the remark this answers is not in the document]
MartinFlores751: It will not compile with `std::string_view`. Can do a `const std::string&`.

core/src/common.cpp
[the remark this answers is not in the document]
alanking: Either way. `info` just seemed a bit much if this is a common occurrence. On the other hand, maybe it's just right. I'm not familiar enough with the feature to know better.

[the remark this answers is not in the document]
trel: Agreed.   Not sure how often someone starts a flow with a bad token.

Being HTTP, I wouldn't be surprised if scripts are always trying shenanigans to get in... so that would suggest `debug`...   but if that's not the case, yes, `info` feels correct enough.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -362,6 +364,434 @@ namespace irods::http
...
+
+		return json_res;
+	}
+
+	/// Fetches JWKs from the location specified by the OpenID Provider
+	///
+	/// See OpenID Connect Discovery 1.0 Section 3 for info on jwks_uri
+	/// See RFC 7517 for more information on JSON Web Key (JWK)
+	///
+	/// \returns A std::string represting the JWKs from the OpenID Provider
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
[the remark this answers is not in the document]
MartinFlores751: Logging as debug.

core/src/common.cpp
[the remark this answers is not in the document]
MartinFlores751: Moved outside the if statements.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

README.md
```diff
@@ -243,6 +243,10 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
                 // Optional unless running as a protected resource.
                 "client_secret": "xxxxxxxxxxxxxxx",
 
+                // The realm secret used validating signed JWT Access Tokens using a symmetric algorithm.
+                // Used if provided, otherwise, defaults to validating with "client_secret".
+                "realm_secret": "xxxxxxxxxxxxxxx",
```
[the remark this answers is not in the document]

core/src/common.cpp
korydraughn: Yeah, let's move them into their own hpp/cpp files then.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+						add_asymmetric_alg_from_jwk(_verifier, _jwk, alg);
+						return;
+					}
+
+					logging::trace("{}: JWK [kty] does not match JWT desired [kty], ignoring.", fn);
+					return;
+				}
+
+				logging::error("{}: Invalid JWK, missing [kty] claim. Ignoring.", fn);
+				return;
```
korydraughn, line 694: This `return` isn't necessary since we're at the end of the function.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+			});
+
+		// Allow for chaining
+		return _verifier;
+	}
```
korydraughn, line 699: `} // add_algorithms_to_verifier`

Please do the same for the other functions.

korydraughn (commented):
Good stuff.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+	auto validate_using_introspection_endpoint(const std::string& _bearer_token) -> std::optional<nlohmann::json>
+	{
+		namespace logging = irods::http::log;
+
+		body_arguments args{{"token", _bearer_token}, {"token_type_hint", "access_token"}};
+
+		auto json_res{hit_introspection_endpoint(url_encode_body(args))};
+
+		// Validate access token
+		if (!json_res.at("active").get<bool>()) {
```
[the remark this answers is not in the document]
MartinFlores751: `active` is the only required item by the introspection endpoint standard: https://www.rfc-editor.org/rfc/rfc7662.html#section-2.2

Optional items will be validated in another PR/issue.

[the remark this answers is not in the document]
korydraughn: Got it.

README.md
```diff
@@ -243,6 +243,11 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
                 // Optional unless running as a protected resource.
                 "client_secret": "xxxxxxxxxxxxxxx",
 
+                // The realm secret used validating signed JWT Access Tokens using a symmetric algorithm.
```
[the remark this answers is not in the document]
MartinFlores751:
```suggestion
                // The secret used when validating a JWT signed using a symmetric algorithm (e.g. HS256).
```
Something like this?

README.md
```diff
@@ -243,6 +243,11 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
...
+                // Used if provided, otherwise, defaults to validating with "client_secret".
+                // If provided, this MUST be base64url encoded.
```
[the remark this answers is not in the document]
MartinFlores751: Should we also have MUST be lowercase?

README.md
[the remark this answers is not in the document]
korydraughn: Yes, that's better.

README.md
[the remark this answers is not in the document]

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
 		return std::nullopt;
 	}
 
+	using jwt_verifier = jwt::verifier<jwt::default_clock, jwt::traits::nlohmann_json>;
+
+	/// Validates an OAuth 2.0 Access Token using the Introspection Endpoint
+	/// See RFC 7662 on OAuth 2.0 Token Introspection for more details
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+
+		// Build Request
+		constexpr auto http_version_number{11};
+		beast::http::request<beast::http::string_body> req{beast::http::verb::get, url.path(), http_version_number};
+		req.set(beast::http::field::host, irods::http::create_host_field(url, *port));
+		req.set(beast::http::field::user_agent, irods::http::version::server_name);
+		req.set(beast::http::field::accept, "application/json");
+		req.prepare_payload();
+
+		// Send request & receive response
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+		return res.body();
+	}
+
+	/// Adds the specified symmetric algorithm \p _alg to the verifier \p _verifier, using the secrets provided in the
+	/// HTTP API configuration
+	///
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in,out] _verifier The jwt_verifier to add additional verification algorithms to.
+	/// \param[in]     _alg      The signing algorithm requested by the signed JWT.
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+		// Use a realm_secret if provided. It should be base64url encoded, as the key might
+		// not be ASCII printable.
+		if (auto realm_secret{irods::http::globals::oidc_configuration().find("realm_secret")};
+		    realm_secret != std::end(irods::http::globals::oidc_configuration()))
+		{
+			key = jwt::base::decode<jwt::alphabet::base64url>(
+				jwt::base::pad<jwt::alphabet::base64url>(realm_secret->get_ref<const std::string&>()));
+		}
+		// While not for access tokens, ID Tokens are signed using the client_secret.
+		// Some OpenID providers may do the same for access tokens
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+	}
+
+	/// Adds the specified asymmetric algorithm \p _alg to the verifier \p _verifier, using the additional information
+	/// provided by the JWK \p _jwk.
+	///
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in,out] _verifier The jwt_verifier to add additional verification algorithms to.
+	/// \param[in]     _jwk      The JWK containing the required JWA information.
+	/// \param[in]     _alg      The signing algorithm requested by the signed JWT.
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+	/// Adds verification algorithm(s) to the \p _verifier based on either the given 'kid' or algorthims matching
+	/// the family of the desired algorithm \p _alg.
+	///
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in,out] _verifier The jwt_verifier to add additional verification algorithms to.
+	/// \param[in]     _jwks     The JWKs to search through.
+	/// \param[in]     _jwt      The decoded JWT that needs to be verified.
+	///
+	/// \returns A reference to the provided jwt_verifier, \p _verifier, allowing for chaining.
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+
+	/// Adds the specified asymmetric algorithm \p _alg to the verifier \p _verifier, using the additional information
+	/// provided by the JWK \p _jwk.
+	///
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in,out] _verifier The jwt_verifier to add additional verification algorithms to.
+	/// \param[in]     _jwk      The JWK containing the required JWA information.
+	/// \param[in]     _alg      The signing algorithm requested by the signed JWT.
+	auto add_asymmetric_alg_from_jwk(
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+	}
+
+	/// Adds the specified symmetric algorithm \p _alg to the verifier \p _verifier, using the secrets provided in the
+	/// HTTP API configuration
+	///
+	/// See RFC 7518 for details on JSON Web Algorithms (JWA)
+	///
+	/// \param[in,out] _verifier The jwt_verifier to add additional verification algorithms to.
+	/// \param[in]     _alg      The signing algorithm requested by the signed JWT.
+	auto add_symmetric_alg(jwt_verifier& _verifier, std::string_view _alg) -> void
```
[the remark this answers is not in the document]
MartinFlores751: Done.

README.md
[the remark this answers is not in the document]
trel: signed using -> signed with ?

README.md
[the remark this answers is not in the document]
trel: lowercase is fine.    feels like you've been deep in those RFCs...

README.md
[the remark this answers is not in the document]

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Done.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+	/// See OpenID Connect Discovery 1.0 Section 3 for info on jwks_uri
+	/// See RFC 7517 for more information on JSON Web Key (JWK)
+	///
+	/// \returns A std::string representing the JWKs from the OpenID Provider
+	auto fetch_jwks_from_openid_provider() -> std::string
+	{
+		namespace logging = irods::http::log;
+
+		const auto jwks_uri{
+			irods::http::globals::oidc_endpoint_configuration().at("jwks_uri").get_ref<const std::string&>()};
```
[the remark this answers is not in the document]
MartinFlores751: Yes, the openid_connect stanza exists, and the openid well-known endpoint should have already been successfully queried.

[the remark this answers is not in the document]
korydraughn: Good. Resolving.

core/src/common.cpp
```diff
@@ -362,6 +364,448 @@ namespace irods::http
...
+	/// \param[in,out] _verifier The jwt_verifier to add additional verification algorithms to.
+	/// \param[in]     _alg      The signing algorithm requested by the signed JWT.
+	auto add_symmetric_alg(jwt_verifier& _verifier, std::string_view _alg) -> void
+	{
+		namespace logging = irods::http::log;
+
+		auto algorithm_family{_alg.substr(0, 2)};
+		auto algorithm_specifics{_alg.substr(2)};
+
+		std::string key;
```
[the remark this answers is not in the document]
MartinFlores751: Done.

core/src/common.cpp
MartinFlores751: Moved.

core/src/common.cpp
MartinFlores751: > What are the pros and cons of letting it bubble up?

The processing of the JWT stops, as it should imo, as the JWK is invalid.
This means there is likely an issue with the OpenID provider.

The cons are that there may be another JWK that is valid?

> What does the admin see?

The admin should see a complaint about an invalid JWK (logs).

> What does the client see?

Client sees an auth failure.

> What should the client do if they provide an invalid JWK?

The client does not provide the JWK, the OpenID Provider does, when querying the well-known endpoint.

korydraughn: > This means there is likely an issue with the OpenID provider.
>
> The cons are that there may be another JWK that is valid?

So if there's an issue with the OIDC provider, then there's no reason to worry about another JWK being valid, correct?

Is this guaranteed to be an issue with the OIDC provider?

core/include/irods/private/http_api/openid.hpp
```diff
@@ -0,0 +1,24 @@
+#ifndef IRODS_HTTP_API_OPENID_HPP
+#define IRODS_HTTP_API_OPENID_HPP
+
+#include <optional>
+#include <string>
```
korydraughn, line 5: Move below boost includes.

core/include/irods/private/http_api/openid.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <boost/beast/http/string_body.hpp>
+#include <boost/url/url_view.hpp>
+
+namespace irods::http::openid
+{
+	auto create_oidc_request(boost::urls::url_view _url)
+		-> boost::beast::http::request<boost::beast::http::string_body>;
+	auto validate_using_introspection_endpoint(const std::string& _bearer_token) -> std::optional<nlohmann::json>;
+	auto validate_using_local_validation(const jwt::decoded_jwt<jwt::traits::nlohmann_json>& _jwt)
+		-> std::optional<nlohmann::json>;
```
korydraughn, line 21: Add empty line between function declarations. Just makes it a little easier to read.

core/src/openid.cpp
```diff
@@ -0,0 +1,544 @@
...
+			const auto& client_id{
+				irods::http::globals::oidc_configuration().at("client_id").get_ref<const std::string&>()};
+			const auto& client_secret{secret_key->get_ref<const std::string&>()};
+			const auto auth_string{fmt::format("Basic {}", format_bearer_token(client_id, client_secret))};
+
+			req.set(beast::http::field::authorization, auth_string);
+		}
+
+		return req;
+	}
```
korydraughn, line 49: Missing trailing comment containing function name.

core/src/openid.cpp
```diff
@@ -0,0 +1,544 @@
...
+		req.prepare_payload();
+
+		// Send request & receive response
+		auto res{tcp_stream->communicate(req)};
+
+		logging::debug("{}: Received the following response: [{}]", __func__, res.body());
+
+		// JSONize response
+		return nlohmann::json::parse(res.body());
+	}
```
korydraughn, line 89: Missing trailing comment containing function name.

korydraughn (commented):
Looking real good.

Please run the test suite with asan enabled.

Is there anything else left to do for this?

core/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ target_link_libraries(
   "${IRODS_EXTERNALS_FULLPATH_BOOST}/lib/libboost_program_options.so"
   "${IRODS_EXTERNALS_FULLPATH_BOOST}/lib/libboost_url.so"
   CURL::libcurl
+  jwt-cpp::jwt-cpp
```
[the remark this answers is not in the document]
korydraughn: We need a new externals package for this, correct?

alanking (commented):
Just saw one thing. Very nice

core/include/irods/private/http_api/openid.hpp
MartinFlores751: Should I also throw the doxygen up there?

core/src/common.cpp
korydraughn: Okay. Letting it ride until we learn more.

Please mark as resolved if you feel there's nothing to address and the implementation is good.

core/include/irods/private/http_api/openid.hpp
MartinFlores751: Done.

core/include/irods/private/http_api/openid.hpp
MartinFlores751: Done.

core/src/openid.cpp
MartinFlores751: Added.

core/src/openid.cpp
MartinFlores751: Added.

core/include/irods/private/http_api/openid.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace irods::http::openid
+{
+	auto create_oidc_request(boost::urls::url_view _url) -> boost::beast::http::request<boost::beast::http::string_body>;
+	auto validate_using_introspection_endpoint(const std::string& _bearer_token) -> std::optional<nlohmann::json>;
+	auto validate_using_local_validation(const jwt::decoded_jwt<jwt::traits::nlohmann_json>& _jwt)
+		-> std::optional<nlohmann::json>;
+} //namespace irods::http::openid
+
+#endif // IRODS_HTTP_API_TRANSPORT_HPP
```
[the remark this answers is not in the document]
MartinFlores751: Fixed.

MartinFlores751:
Going to rebase onto latest.

core/CMakeLists.txt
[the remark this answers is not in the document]
MartinFlores751: No, it's just updating the existing one.

core/include/irods/private/http_api/openid.hpp
korydraughn: > Should I also throw the doxygen up there?

Sure. That will only help us in the future.

core/src/common.cpp
MartinFlores751: Resolving.

core/CMakeLists.txt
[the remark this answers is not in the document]
korydraughn: Please correct me if I'm wrong, but doesn't this require a version bump of the jwt lib?

core/include/irods/private/http_api/openid.hpp
```diff
@@ -14,7 +14,7 @@
 
 namespace irods::http::openid
 {
-	auto create_oidc_request(boost::urls::url_view _url)
+	auto create_oidc_request(const boost::urls::url_view& _url)
```
alanking, line 17: According to the Boost docs, `url_view` functions like a `string_view`. My understanding is that passing by value is favored for `string_view`s over passing by reference. Maybe it doesn't matter so much since this is a "private" function and doesn't need to be particularly optimized...

alanking: For reference: https://www.boost.org/doc/libs/1_83_0/libs/url/doc/html/url/ref/boost__urls__url_view.html#url.ref.boost__urls__url_view.description

MartinFlores751:
Ran general tests:
```
$ python3 test_irods_http_api.py 
s.......s...........................s...........s...................s...ss...ss..sss..sss...s.s.....
Ran 100 tests in 20.165s

OK (skipped=17)
```

Ran auth and general tests (client mode):
```
$ pytest
...
=== 101 passed, 18 skipped in 21.04s ===
...
```

Ran auth and general tests (protected_resource mode):
```
$ pytest
...
=== 97 passed, 22 skipped in 21.31s ===
...
```

---

Still need to test other bits (all supported signing algs) manually.

core/include/irods/private/http_api/openid.hpp
MartinFlores751: Yeah, I agree with you to treat this as a string_view. Just added this change to address clang-tidy.

Revert and add lint ignore?

korydraughn: Yes please.

MartinFlores751: Done.

korydraughn:
Do those test runs include the use of asan? If not, that's fine. Can rerun the tests with asan as the final step before approving.

MartinFlores751:
> Do those test runs include the use of asan? If not, that's fine. Can rerun the tests with asan as the final step before approving.

I ran CMake with `option(IRODS_ENABLE_ADDRESS_SANITIZER "Enables detection of memory leaks and other features provided by Address Sanitizer." ON)`, so it should be on.

MartinFlores751:
Reran tests after revert, all pass in same manner.

Doing manual tests now.

trel:
exciting.   go go go.

MartinFlores751:
Tested all supported algorithms work, tested expired JWT, tested JWK missing from JWKs. All work.

trel:
Sounds like squashing has entered the conversation.   One more vote for that please…

MartinFlores751:
Bonus tests invalid JWTs, general garbage:

command
```
$ curl -H "Authorization: Bearer abc.def.ghi" --data-urlencode 'op=stat' --data-urlencode 'lpath=/tempZone' -G http://172.18.0.3:9000/irods-http-api/0.4.0/collections -v
*   Trying 172.18.0.3:9000...
* Connected to 172.18.0.3 (172.18.0.3) port 9000
* using HTTP/1.x
> GET /irods-http-api/0.4.0/collections?op=stat&lpath=%2ftempZone HTTP/1.1
> Host: 172.18.0.3:9000
> User-Agent: curl/8.10.1
> Accept: */*
> Authorization: Bearer abc.def.ghi
> 
* Request completely sent off
< HTTP/1.1 401 Unauthorized
< Server: irods_http_api/0.4.0 (76555f99cfad6a99b7d6a9b1099806840ee87022)
< Content-Type: application/json
< Content-Length: 0
< 
* Connection #0 to host 172.18.0.3 left intact
```

log
```
[2024-11-05 18:15:28.091] [P:14299] [debug] [T:14299] resolve_client_identity: Authorization value: [Bearer abc.def.ghi]
[2024-11-05 18:15:28.091] [P:14299] [debug] [T:14299] resolve_client_identity: Bearer token: [abc.def.ghi]
[2024-11-05 18:15:28.091] [P:14299] [debug] [T:14299] resolve_client_identity: [json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - invalid literal; last read: 'i'
[2024-11-05 18:15:28.093] [P:14299] [debug] [T:14299] hit_introspection_endpoint: Received the following response: [{"active":false}]
[2024-11-05 18:15:28.093] [P:14299] [warning] [T:14299] validate_using_introspection_endpoint: Access token is invalid or expired.
[2024-11-05 18:15:28.093] [P:14299] [error] [T:14299] resolve_client_identity: Could not find bearer token matching [abc.def.ghi].
```

---

command
```
curl -H 'Authorization: Bearer {"hello": "world"}' --data-urlencode 'op=stat' --data-urlencode 'lpath=/tempZone' -G http://172.18.0.3:9000/irods-http-api/0.4.0/collections -v
*   Trying 172.18.0.3:9000...
* Connected to 172.18.0.3 (172.18.0.3) port 9000
* using HTTP/1.x
> GET /irods-http-api/0.4.0/collections?op=stat&lpath=%2ftempZone HTTP/1.1
> Host: 172.18.0.3:9000
> User-Agent: curl/8.10.1
> Accept: */*
> Authorization: Bearer {"hello": "world"}
> 
* Request completely sent off
< HTTP/1.1 401 Unauthorized
< Server: irods_http_api/0.4.0 (76555f99cfad6a99b7d6a9b1099806840ee87022)
< Content-Type: application/json
< Content-Length: 0
< 
* Connection #0 to host 172.18.0.3 left intact
```

log
```
[2024-11-05 18:19:54.527] [P:14299] [debug] [T:14299] resolve_client_identity: Authorization value: [Bearer {"hello": "world"}]
[2024-11-05 18:19:54.527] [P:14299] [debug] [T:14299] resolve_client_identity: Bearer token: [{"hello": "world"}]
[2024-11-05 18:19:54.527] [P:14299] [debug] [T:14299] resolve_client_identity: invalid token supplied
[2024-11-05 18:19:54.528] [P:14299] [debug] [T:14299] hit_introspection_endpoint: Received the following response: [{"active":false}]
[2024-11-05 18:19:54.528] [P:14299] [warning] [T:14299] validate_using_introspection_endpoint: Access token is invalid or expired.
[2024-11-05 18:19:54.528] [P:14299] [error] [T:14299] resolve_client_identity: Could not find bearer token matching [{"hello": "world"}].
```

korydraughn:
Please capture that in an existing or new issue for test automation. You can link it to this PR.

core/CMakeLists.txt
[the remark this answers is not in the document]
MartinFlores751: I think I get you now. New version means new package? If that's the case, then yes.

MartinFlores751:
No other tests unless others suggested. Should be ready otherwise.

korydraughn:
So you feel this PR is complete and ready for squashing?

MartinFlores751:
Yes, the PR should be complete.

korydraughn:
Squash it, no pounds.

Open a PR in the externals repo and bump the version for the jwt package.

MartinFlores751:
Squashed.

README.md
```diff
@@ -243,6 +243,11 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
                 // Optional unless running as a protected resource.
                 "client_secret": "xxxxxxxxxxxxxxx",
 
+                // The secret used when validating a JWT signed with a symmetric algorithm (e.g. HS256).
+                // Used if provided, otherwise, defaults to validating with "client_secret".
+                // If provided, it MUST be base64url encoded.
```
korydraughn, line 248: The line length causes a horizontal scrollbar to appear. Please adjust.

I noticed the `user_mapping` stanza causes that too. Please adjust that as well.

korydraughn:
The jwt-cpp reference in the top-level CMakeLists.txt file will need its version number updated once you find a stable version that satisfies the requirements.

MartinFlores751:
Also need to update the Docker build file to have the latest version of jwt-cpp.

core/CMakeLists.txt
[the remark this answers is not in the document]
MartinFlores751: Externals updated.

korydraughn (commented):
Squash it.

MartinFlores751:
squashed.

korydraughn:
Once the externals package is published, please build the docker images/containers and confirm it works as intended.

We'll get the release notes updated and release it after that.

korydraughn:
Oops, forgot about PR #358. So, there's still more work to be done after this PR.

MartinFlores751:
Externals built, the http api images build with the package as well as this pr.

alanking (approved):
Numbers all look right to me. Please await one more approval before #'ing.

core/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ add_library(
   "${CMAKE_CURRENT_SOURCE_DIR}/src/process_stash.cpp"
   "${CMAKE_CURRENT_SOURCE_DIR}/src/session.cpp"
   "${CMAKE_CURRENT_SOURCE_DIR}/src/transport.cpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/src/openid.cpp"
```
korydraughn, line 20: Move openid.cpp file between multipart_form_data.cpp and process_stash.cpp

korydraughn (approved):
Looks good.

Once you handle the ordering of the cpp files, squash and pound it.

core/CMakeLists.txt
MartinFlores751: Done.

MartinFlores751:
Squashed.

alanking (approved):
cpp file ordering looks right. # it

MartinFlores751:
Pounded.

Issue #269, opened 2024-03-15
collapse version information to just CMakeLists.txt if possible


trel:
## Feature

In addition to CMakeLists.txt - there are three places where the version number has to be kept in sync.

We'd like to eliminate these additional locations.

https://github.com/irods/irods_client_http_api/pull/268#issuecomment-2000461791

PR #347, opened 2024-10-31
[#269] Collapse version information to single point of control

korydraughn:
- Schema files/info does not need to include version info
- Config file for testing does not need to include version info

### DIFF
--- a/core/src/main.cpp
+++ b/core/src/main.cpp
@@ -163,7 +163,7 @@ constexpr auto default_jsonschema() -> std::string_view
 	// clang-format on
 	return R"({{
     "$schema": "https://json-schema.org/draft/2020-12/schema",
-    "$id": "https://schemas.irods.org/irods-http-api/0.4.0/schema.json",
+    "$id": "https://schemas.irods.org/irods-http-api/config.json",
     "type": "object",
     "properties": {{
         "http_server": {{
--- a/test/config.py
+++ b/test/config.py
@@ -6,7 +6,7 @@ test_config = {
 
     'host': 'localhost',
     'port': 9000,
-    'url_base': '/irods-http-api/0.4.0',
+    'url_base': '/irods-http-api/<version>',
 
     'openid_connect': {
         'mode': 'client'
@@ -30,7 +30,7 @@ test_config = {
 
 schema = {
     '$schema': 'http://json-schema.org/draft-07/schema#',
-    '$id': 'https://schemas.irods.org/irods-http-api/test/0.4.0/test-schema.json',
+    '$id': 'https://schemas.irods.org/irods-http-api/test_config.json',
     'type': 'object',
     'properties': {
         'host': {


test/config.py
```diff
@@ -6,7 +6,7 @@
 
     'host': 'localhost',
     'port': 9000,
-    'url_base': '/irods-http-api/0.4.0',
+    'url_base': '/irods-http-api/<version>',
```
MartinFlores751, line 9: Don't forget to update the tests to populate the version number through some means. Otherwise, we won't be able to test the endpoints.

MartinFlores751 (commented):
One comment.

core/src/main.cpp
```diff
@@ -161,7 +161,7 @@ constexpr auto default_jsonschema() -> std::string_view
 	// clang-format on
 	return R"({{
     "$schema": "https://json-schema.org/draft/2020-12/schema",
-    "$id": "https://schemas.irods.org/irods-http-api/0.4.0/schema.json",
+    "$id": "https://schemas.irods.org/irods-http-api/config.json",
```
trel, line 164: neither of these URLs (before or after) are real...

and not sure we want to introduce new dependencies on schemas.irods.org... it is still alive only because early 4.2 pointed there to resolve things I believe...

test/config.py
```diff
@@ -30,7 +30,7 @@
 
 schema = {
     '$schema': 'http://json-schema.org/draft-07/schema#',
-    '$id': 'https://schemas.irods.org/irods-http-api/test/0.4.0/test-schema.json',
+    '$id': 'https://schemas.irods.org/irods-http-api/test_config.json',
```
trel, line 33: same here - might need some more ... thought.

test/config.py
korydraughn: Not sure what you mean. It is the dev's responsibility to properly configure the test config file before launching the tests. Having the build system update the version was a convenience, not a requirement.

core/src/main.cpp
korydraughn: Correct. The URL is only an identifier. It means nothing to the application. iRODS 5 does the same thing.

This aligns with the issue about removing relative paths from the schema files. See https://github.com/irods/irods/issues/6283.

trel: okay, so... do we have a suitable suggestion for replacement here?

is the `$id` required?   and what can satisfy it?

korydraughn: Replacement? Can you elaborate?

The requirements of `$id` are already satisfied. The only requirement is that it be unique. It does not need to resolve to anything.

trel: unique... so it doesn't even need to be a url?   

suggestion is... no.

https://json-schema.org/understanding-json-schema/basics#declaring-a-unique-identifier

trel: the latest draft...
https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#section-8.2.1

> Note that this URI is an identifier and not necessarily a network locator. In the case of a network-addressable URL, a schema need not be downloadable from its canonical URI.

so we're good with
```
https://schemas.irods.org/irods-http-api/config.json 
```

marking as resolved.

korydraughn:
This is ready for final review.

alanking (commented):
So just to be clear on the test/config.py change... The way that the tests are run by the uninitiated (e.g. me) remains unchanged?

korydraughn:
First time users will need to update the test/config.py file to point to the correct location of the HTTP API (i.e. replace `<version>`).

Other than that, running the tests is as it's always been.

trel:
if we have a 'how to run tests' area, please make sure it says that.   otherwise, we're good.

korydraughn:
We don't currently have a section on running tests.

I'll open an issue for that. Can handle that at a later time.

alanking (approved):
Please link new issue and then # it

korydraughn:
Created #350 for documenting the process of running tests.

korydraughn:
#'d

Issue #337, opened 2024-09-26
Running the HTTP API with a configuration file that does not exist produces an unclear error


MartinFlores751:
Running the HTTP API with a configuration file that does not exist results in:
`Error: [json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal`

Example line to execute: `irods_http_api /this_file_does_not_exist.json`

The HTTP API should log an error stating that the provided configuration does not exist.

PR #348, opened 2024-10-31
[#337] Print clear error message when config file path is invalid

korydraughn:
Here's an example:
```
$ ./irods_http_api ../irods_client_http_api/config.jso
Error: Cannot access configuration file [../irods_client_http_api/config.jso].
```

### DIFF
--- a/core/src/main.cpp
+++ b/core/src/main.cpp
@@ -41,6 +41,7 @@
 #include <csignal>
 #include <cstdint>
 #include <cstdlib>
+#include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <iterator>
@@ -74,6 +75,7 @@ extern "C" const char* __asan_default_options()
 namespace beast   = boost::beast; // from <boost/beast.hpp>
 namespace net     = boost::asio;  // from <boost/asio.hpp>
 namespace po      = boost::program_options;
+namespace fs      = std::filesystem;
 namespace logging = irods::http::log;
 
 using json = nlohmann::json;
@@ -946,11 +948,25 @@ auto main(int _argc, char* _argv[]) -> int
 		}
 
 		if (vm.count("config-file") == 0) {
-			fmt::print(stderr, "Error: Missing [CONFIG_FILE_PATH] parameter.");
+			fmt::print(stderr, "Error: Missing [CONFIG_FILE_PATH] parameter.\n");
 			return 1;
 		}
 
-		const auto config = json::parse(std::ifstream{vm["config-file"].as<std::string>()});
+		const auto& config_file_path = vm["config-file"].as<std::string>();
+		if (!fs::exists(config_file_path)) {
+			fmt::print(stderr, "Error: Configuration file [{}] does not exist.\n", config_file_path);
+			return 1;
+		}
+
+		const auto config = [&config_file_path] {
+			std::ifstream file{config_file_path};
+			if (!file.is_open()) {
+				throw std::runtime_error{fmt::format("Cannot open configuration file [{}].", config_file_path)};
+			}
+
+			return json::parse(file);
+		}();
+
 		irods::http::globals::set_configuration(config);
 
 		{


core/src/main.cpp
```diff
@@ -946,11 +958,17 @@ auto main(int _argc, char* _argv[]) -> int
 		}
 
 		if (vm.count("config-file") == 0) {
-			fmt::print(stderr, "Error: Missing [CONFIG_FILE_PATH] parameter.");
+			fmt::print(stderr, "Error: Missing [CONFIG_FILE_PATH] parameter.\n");
```
MartinFlores751, line 961: Do we have `fmt::println(...)`?

core/src/main.cpp
```diff
@@ -53,6 +53,12 @@
 #include <utility>
 #include <vector>
 
+#ifdef __cpp_lib_filesystem
+#  include <filesystem>
+#else
+#  include <boost/filesystem.hpp>
+#endif
```
alanking, line 60: Just wondering: What situation would not have `std::filesystem` available for this project?

core/src/main.cpp
```diff
@@ -946,11 +958,17 @@ auto main(int _argc, char* _argv[]) -> int
 		}
 
 		if (vm.count("config-file") == 0) {
-			fmt::print(stderr, "Error: Missing [CONFIG_FILE_PATH] parameter.");
+			fmt::print(stderr, "Error: Missing [CONFIG_FILE_PATH] parameter.\n");
+			return 1;
+		}
+
+		const auto config_file_path = vm["config-file"].as<std::string>();
```
MartinFlores751, line 965: Can we get this as a reference?

MartinFlores751 (commented):
Two comments.

alanking (dismissed):
Seems fine to me. Had one question, but I think you can go ahead and # it if you feel it's ready.

alanking:
Oh, nevermind - don't # til @MartinFlores751 approves. :)

core/src/main.cpp
MartinFlores751: Curious about this too. If we do need this, should this be done in the user mapping plugin as well for `local_file.cpp`?

core/src/main.cpp
korydraughn: I don't think the version of fmtlib we use supports that version of print.

_I wish it did, but oh well._

core/src/main.cpp
korydraughn: I don't think so?

I've never seen any boost program options example demonstrating acquiring a reference. If you can find one, that would be great.

MartinFlores751: Skimming through the boost filesystem docs, it looks like what you call takes a path ref, and the path ref constructor takes a ref to the string? You can ignore this comment if you don't want to verify, one string copy isn't bad: https://www.boost.org/doc/libs/1_81_0/libs/filesystem/doc/reference.html

core/src/main.cpp
korydraughn: I think this is actually unnecessary now.

Confirmed via compiler explorer. Our compiler is new enough to not need this.

Will update.

core/src/main.cpp
korydraughn: This is good. Apparently, `.as()` returns a reference to the contents of a `boost::any` object. So I think we're good with just appending an ampersand to `auto`.

Will test that real quick.

core/src/main.cpp
```diff
@@ -946,11 +948,17 @@ auto main(int _argc, char* _argv[]) -> int
...
 
 		if (vm.count("config-file") == 0) {
-			fmt::print(stderr, "Error: Missing [CONFIG_FILE_PATH] parameter.");
+			fmt::print(stderr, "Error: Missing [CONFIG_FILE_PATH] parameter.\n");
 			return 1;
 		}
 
-		const auto config = json::parse(std::ifstream{vm["config-file"].as<std::string>()});
+		const auto& config_file_path = vm["config-file"].as<std::string>();
+		if (!fs::is_regular_file(config_file_path)) {
```
trel, line 956: is_regular_file will check if it can be opened?   considering it exists, but has bad permissions...

korydraughn: No. It confirms the file is NOT something other than `regular`. See https://en.cppreference.com/w/cpp/filesystem/file_type for the various file types that can be returned.

Hmm, but now I'm thinking it should just be an existence check since it distinguishes symlinks.

core/src/main.cpp
korydraughn: Confirmed that `fs::path` cannot be used. I did not investigate why. Everything I tried resulted in Boost.Any complaining.

Will continue to use `auto&` and `std::string`.

core/src/main.cpp
trel: so if the file exists, and is regular, but cannot be read because of permissions...

does this catch that?  or we handle that later when it tries to read/parse it?

korydraughn: Confirmed the changes do not cover permissions. If the application cannot read the file, you get the same error that was reported in the issue.

How common is it for applications to check permissions for config files though? Don't think I've ever encountered a situation with that before.

trel: ah, right... it will just say, cannot access... which is correct.    okay, so an existence check is probably sufficient here.

korydraughn: Updated code to report an error if the file does not exist or cannot be opened.

Done.

korydraughn:
This is what you get when the file does not exist.
```
Error: Configuration file [../irods_client_http_api/config.j] does not exist.
```

This is what you get when the file exists, but cannot be read (perhaps due to permissions).
```
Error: Cannot open configuration file [../irods_client_http_api/config.json].
```

alanking (commented):
I think this is ready for squashing.

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
#'d

Issue #270, opened 2024-03-18
protected_resource bearer token validation 


ll4strw:
I am wondering if it is desirable to add extended checks on a response obtained as a result of a token introspection request. The code below seems to grant access to protected resources as long as a bearer token is still active and its associated user exists in iRODS:

https://github.com/irods/irods_client_http_api/blob/48b63758b056e11008a6b1e876f0ed427a3f39c2/core/src/common.cpp#L402-L429

What happens if a valid and active bearer token meant for another app/client, but for a valid iRODS user, is presented to the API server? Would this be let in? Could the token validation be performed also on the response  `client_id` and `aud`? For instance, I could imagine an extra configuration parameter called `audience` to add an extra level of security

```
std::string aud = irods::http::globals::oidc_configuration().at("audience").get_ref<const std::string&>();
if (json_res.at("aud").get<string>() != aud) {
	logging::warn("{}: Access token is not meant for me.", __func__);
	return {.response = fail(status_type::unauthorized)};
}
```
Thanks in advance, L.

PR #358, opened 2024-11-07
[#270] Validate OAuth Token Introspection response

MartinFlores751:
This PR is built on top of #344.

This PR performs further validation of the OAuth introspection response by using the optional response values if available.

Still needs to be tested.

### DIFF
--- a/core/src/openid.cpp
+++ b/core/src/openid.cpp
@@ -98,14 +98,78 @@ namespace irods::http::openid
 	{
 		namespace logging = irods::http::log;
 
-		body_arguments args{{"token", _bearer_token}, {"token_type_hint", "access_token"}};
+		const body_arguments args{{"token", _bearer_token}, {"token_type_hint", "access_token"}};
 
 		auto json_res{hit_introspection_endpoint(url_encode_body(args))};
 
-		// Validate access token
-		if (!json_res.at("active").get<bool>()) {
+		// Check that the access token is active. If active is false or missing from
+		// the response, reject the token.
+		// The exact meaning of an active state may vary between OpenID Providers,
+		// so we should assert the other attributes when available.
+		if (const auto active_iter{json_res.find("active")};
+		    active_iter == std::end(json_res) || !active_iter->get<bool>()) {
 			logging::warn("{}: Access token is invalid or expired.", __func__);
 			return std::nullopt;
+		}
+
+		// The rest of the members are optional, but validate the ones that do exist.
+		// In total, there are 11 optional members defined by the RFC that may be used.
+
+		// Token must not be used early if given 'not before' time.
+		if (auto nbf_iter{json_res.find("nbf")}; nbf_iter != std::end(json_res)) {
+			logging::trace("{}: Attempting [nbf] validation.", __func__);
+
+			// Get current time
+			const auto now{std::chrono::system_clock::now()};
+
+			// Convert to appropriate representation
+			const std::chrono::seconds nbf_seconds{nbf_iter->get<std::time_t>()};
+
+			// Setup expire time
+			const std::chrono::time_point<std::chrono::system_clock> not_before_time{nbf_seconds};
+
+			// Reject tokens that are used too early
+			if (now < not_before_time) {
+				logging::warn("{}: Current time is before [nbf]. Validation failed.", __func__);
+				return std::nullopt;
+			}
+		}
+
+		// We should be part of the intended audience for the access token.
+		if (const auto aud_iter{json_res.find("aud")}; aud_iter != std::end(json_res)) {
+			logging::trace("{}: Attempting [aud] validation.", __func__);
+
+			const auto& client_id{irods::http::globals::oidc_configuration().at("client_id")};
+			const auto& aud{*aud_iter};
+
+			if (aud.is_array()) {
+				auto aud_list_iter{aud.items()};
+				auto client_id_in_list{std::any_of(
+					std::begin(aud_list_iter), std::end(aud_list_iter), [&client_id](const auto& _kvp) -> bool {
+						return _kvp.value() == client_id;
+					})};
+
+				// If we are not in the list, reject the token
+				if (!client_id_in_list) {
+					logging::warn("{}: Could not find our [client_id] in [aud]. Validation failed.", __func__);
+					return std::nullopt;
+				}
+			}
+			else if (aud.get_ref<const std::string&>() != client_id) {
+				logging::warn("{}: Could not find our [client_id] in [aud]. Validation failed.", __func__);
+				return std::nullopt;
+			}
+		}
+
+		// The 'iss' provided should match the 'issuer' retrieved from the OpenID Provider's
+		// well-known endpoint.
+		if (const auto iss_iter{json_res.find("iss")}; iss_iter != std::end(json_res)) {
+			logging::trace("{}: Attempting [iss] validation.", __func__);
+			if (iss_iter->get_ref<const std::string&>() !=
+			    irods::http::globals::oidc_endpoint_configuration().at("issuer")) {
+				logging::warn("{}: [iss] did not match OpenID Provider's [issuer]. Validation failed.", __func__);
+				return std::nullopt;
+			}
 		}
 
 		return json_res;


core/src/openid.cpp
```diff
@@ -102,12 +102,100 @@ namespace irods::http::openid
...
 			return std::nullopt;
 		}
 
+		// The rest of the members are optional, but validate the ones that do exist.
+		// TODO: Verify that this is actually how this works. I don't think we can check this 'client_id'.
+		if (auto client_id_iter{json_res.find("client_id")}; client_id_iter != std::end(json_res)) {
+			if (client_id_iter->get_ref<const std::string&>() != irods::http::globals::oidc_configuration().at("client_id")) {
+				return std::nullopt;
+			}
+		}
```
MartinFlores751, line 119: I don't think `client_id` refers to us.

core/src/openid.cpp
```diff
@@ -102,12 +102,100 @@ namespace irods::http::openid
...
+		// While not required, ensure the token type is something we can use.
+		// TODO: Consider removing this part. Knowing the 'token_type' after the server has already
+		// processed the token doesn't do anything for us.
+		if (auto token_type_iter{json_res.find("token_type")}; token_type_iter != std::end(json_res)) {
+
+			// There are only 'bearer' tokens and 'mac' tokens defined
+			if (token_type_iter->get_ref<const std::string&>() != "bearer") {
+				return std::nullopt;
+			}
+		}
```
MartinFlores751, line 130: Since we have already 'processed' the token, I don't think knowing whether this a 'bearer' or a 'mac' based token is of interest. We also only process bearer tokens.

core/src/openid.cpp
```diff
@@ -102,12 +102,100 @@ namespace irods::http::openid
...
+			std::chrono::seconds nbf_seconds{nbf_iter->get<int>()};
+
+			// Setup expire time
+			std::chrono::time_point<std::chrono::system_clock> not_before_time{nbf_seconds};
+
+			// Reject tokens that are used too early
+			if (now < not_before_time) {
+				return std::nullopt;
+			}
+		}
```
MartinFlores751, line 181: Processing of `exp`, `iat`, and `nbf` should not be required. The OpenID Provider should validate the token for us (e.g. ensuring token is not used before `nbf`) and represent the validity of a token using the `active` member.

core/src/openid.cpp
```diff
@@ -102,12 +102,100 @@ namespace irods::http::openid
...
+
+		// We should be part of the intended audience for the access token.
+		if (auto aud_iter{json_res.find("aud")}; aud_iter != std::end(json_res)) {
+			// TODO: This can be a list, so we need to check if it is, and found ourselves if it is
+			if (aud_iter->get_ref<const std::string&>() != irods::http::globals::oidc_configuration().at("client_id")) {
+				return std::nullopt;
+			}
+		}
```
MartinFlores751, line 189: Use nlohmann json to check if the response is a list or a string.

Process accordingly.

core/src/openid.cpp
```diff
@@ -102,12 +102,100 @@ namespace irods::http::openid
...
+
+		// The 'iss' provided should match the 'issuer' retrieved from the OpenID Provider's
+		// well-known endpoint.
+		if (auto iss_iter{json_res.find("iss")}; iss_iter != std::end(json_res)) {
+			if (iss_iter->get_ref<const std::string&>() != irods::http::globals::oidc_endpoint_configuration().at("issuer")) {
+				return std::nullopt;
+			}
+		}
```
MartinFlores751, line 197: This may be optional, but the check is fine to leave in.

MartinFlores751 (commented):
Some thoughts on the members checked.

core/src/openid.cpp
MartinFlores751, line 1: Add some logging.

core/src/openid.cpp
MartinFlores751: Removed.

core/src/openid.cpp
MartinFlores751: Removed.

core/src/openid.cpp
MartinFlores751: Added some logs, will improve later.

core/src/openid.cpp
MartinFlores751: Done.

core/src/openid.cpp
MartinFlores751: Doesn't refer to us. Removed check.

core/src/openid.cpp
MartinFlores751: Leaving in.

core/src/openid.cpp
MartinFlores751: Changed the logs.

core/src/openid.cpp
```diff
@@ -98,16 +98,79 @@ namespace irods::http::openid
...
-		// Validate access token
-		if (!json_res.at("active").get<bool>()) {
+		// Check that the access token is active.
+		// The exact meaninng of an active state may vary between OpenID Providers,
+		// so we should assert the other attributes when available.
+		if (const auto active_iter{json_res.find("active")};
+		    active_iter == std::end(json_res) || !active_iter->get<bool>()) {
 			logging::warn("{}: Access token is invalid or expired.", __func__);
 			return std::nullopt;
 		}
```
alanking, line 112: Firstly...
```suggestion
		// Check that the access token is active.
		// The exact meaning of an active state may vary between OpenID Providers,
		// so we should assert the other attributes when available.
		if (const auto active_iter{json_res.find("active")};
		    active_iter == std::end(json_res) || !active_iter->get<bool>()) {
			logging::warn("{}: Access token is invalid or expired.", __func__);
			return std::nullopt;
		}
```

Secondly... Does this mean that we do *not* want to `return` before checking the "other attributes"? The comment suggests to me that an OpenID Provider not providing an "active" attribute or if the "active" attribute is `false` does not necessarily indicate that the token is *not* active. Am I reading that wrong?

core/src/openid.cpp
```diff
@@ -98,16 +98,79 @@ namespace irods::http::openid
...
 
-		body_arguments args{{"token", _bearer_token}, {"token_type_hint", "access_token"}};
+		const body_arguments args{{"token", _bearer_token}, {"token_type_hint", "access_token"}};
 
 		auto json_res{hit_introspection_endpoint(url_encode_body(args))};
 
-		// Validate access token
-		if (!json_res.at("active").get<bool>()) {
+		// Check that the access token is active.
+		// The exact meaninng of an active state may vary between OpenID Providers,
```
korydraughn, line 106: meaninng -> meaning

korydraughn (commented):
Looks correct.

Have you seen the optional code blocks get invoked during testing?

alanking (commented):
Everything looks good except I got a little confused by the comment on the "active" attribute. I think it might need to be reworded, or at least explained for me. :)

core/src/openid.cpp
MartinFlores751: A response with the `active` member set to `false` means that the token is completely invalid (e.g., expired, not recognized). Additionally, no other members are returned in this case.

When a response has `active` as `true`, this means that the provided access token is valid by some measure that the OpenID Provider specifies as active. There are general guarantees, but the specifics are implementation defined.

Quote from the RFC:
> The specifics of a token’s "active" state will vary depending on the implementation of the authorization server and the information it keeps about its tokens, but a "true" value return for the "active" property will generally indicate that a given token has been issued by this authorization server, has not been revoked by the resource owner, and is within its given time window of validity (e.g., after its issuance time and before its expiration time).

Link to response info: https://www.rfc-editor.org/rfc/rfc7662.html#section-2.2

Hopefully, that explains the weird phrasing.

core/src/openid.cpp
MartinFlores751: Done.

MartinFlores751:
> Looks correct.
> 
> Have you seen the optional code blocks get invoked during testing?

Yep. The optional code executes (and doesn't when not present) just fine.

core/src/openid.cpp
alanking: Okay, I see. So, as a general matter, the token *at least* must be "active" == `true`. Beyond that, some providers will have additional checks that must be considered below. The way it's phrased now made me think that some providers might not set "active" but *would* set the other attributes.

MartinFlores751: So, something closer to this?
`// The exact meaning of the state member being true may vary between OpenID Providers...`

MartinFlores751: Added more words.

alanking (commented):
Excellent. Thanks for adding those words. Hopefully it helps someone other than just me. :)

korydraughn:
Fix for clang-tidy workflow incoming!

korydraughn:
The reusable workflow file has been updated, but it seems this PR is still tied to the previous version of the workflow file even though it points to `@main`.

Ignore the clang-tidy.

MartinFlores751:
Ran it myself:
```console
$ git diff -U0 4d0a5d6 | python3 /opt/irods-externals/clang13.0.1-0/share/clang/clang-tidy-diff.py -path /build/irods_http_api_build/ -p1 


193816 warnings generated.
Suppressed 193916 warnings (193814 in non-user code, 2 due to line filter, 100 NOLINT).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.


```

korydraughn:
Good. Looks like it's clean.

If everything is working, squash it.

MartinFlores751:
Squashed.

trel (commented):
i like it.

korydraughn (approved):
Pound it.

MartinFlores751:
Pounded.

Issue #365, opened 2024-11-12
Leave space for the null-terminating byte when using `std::string::copy`


korydraughn:
The following lines need to subtract 1 for the null byte. That guarantees the destination buffer is always null-terminated.
https://github.com/irods/irods_client_http_api/blob/ea07a01194680d89f4c435a3a782f20e9c9aa718/endpoints/authentication/src/main.cpp#L486-L487
https://github.com/irods/irods_client_http_api/blob/ea07a01194680d89f4c435a3a782f20e9c9aa718/endpoints/authentication/src/main.cpp#L493

PR #366, opened 2024-11-12
[#365] Leave space in buffer for null-terminating byte when using `std::string::copy`

korydraughn:
This change is safe as it just reduces the number writable bytes to the destination buffer by 1 as defined by the `CheckAuthCredentialsInput` input struct.

See https://docs.irods.org/4.3.3/doxygen/structCheckAuthCredentialsInput.html.

### DIFF
--- a/endpoints/authentication/src/main.cpp
+++ b/endpoints/authentication/src/main.cpp
@@ -482,14 +482,14 @@ namespace irods::http::handler
 							// credentials for native authentication.
 
 							CheckAuthCredentialsInput input{};
-							username.copy(input.username, sizeof(CheckAuthCredentialsInput::username));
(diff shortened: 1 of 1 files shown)


trel:
was this found by inspection?  or did a tool point you to this?   do we need a comment in the code?

korydraughn:
It was upon inspection. No comment needed as this is just how C/C++ string processing works.

trel:
gotcha.

korydraughn:
And because of similar code in the iRODS 5 PR.

alanking (approved):
Seems good to me. # it

korydraughn:
Tested against iRODS 5 PR and auth still works, so we're good.

Pounded.

PR #370, opened 2024-11-15
[#382] Add build action for PRs

MartinFlores751:
Ccache can be removed if we want a clean build every run.

### DIFF
--- a/.github/workflows/build-irods-http-api-ubuntu.yml
+++ b/.github/workflows/build-irods-http-api-ubuntu.yml
@@ -1,0 +1,84 @@
+name: Build Package
+on: [pull_request]
+jobs:
+  build:
+    env:
+      DEBIAN_FRONTEND: noninteractive
+    name: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ 'ubuntu:20.04', 'ubuntu:22.04', 'ubuntu:24.04' ]
+        irods-version: [ 4.3.2, 4.3.3 ]
+        exclude:
+          - os: 'ubuntu:24.04'
+            irods-version: 4.3.2
+          - os: 'ubuntu:22.04'
+            irods-version: 4.3.3
+          - os: 'ubuntu:20.04'
+            irods-version: 4.3.3
+    runs-on: ubuntu-latest
+    container: ${{ matrix.os }}
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Install Prerequisites
+        if: ${{ matrix.os != 'ubuntu:20.04' }}
+        run:  |
+              apt-get update -qq
+              apt-get install -qq \
+                apt-transport-https \
+                gnupg \
+                lsb-release \
+                wget \
+                g++-11 \
+                gcc-11 \
+                libcurl4-gnutls-dev \
+                libssl-dev \
+                libssl3 \
+                ninja-build
+      - name: Install Prerequisites - Ubuntu 20.04
+        if: ${{ matrix.os == 'ubuntu:20.04' }}
+        run:  |
+              apt-get update -qq
+              apt-get install -qq \
+                apt-transport-https \
+                gnupg \
+                lsb-release \
+                wget \
+                g++-10 \
+                gcc-10 \
+                libcurl4-gnutls-dev \
+                libssl-dev \
+                libssl1.1 \
+                ninja-build
+      - name: Install iRODS
+        run:  |
+              wget -qO - https://packages.irods.org/irods-signing-key.asc | apt-key add - &&
+              echo "deb [arch=amd64] https://packages.irods.org/apt/ $(lsb_release -sc) main" | tee /etc/apt/sources.list.d/renci-irods.list
+              apt-get update -qq
+              apt-get install -qq \
+                "irods-dev=${{ matrix.irods-version }}-0~$(lsb_release -sc)" \
+                "irods-runtime=${{ matrix.irods-version }}-0~$(lsb_release -sc)"
+      - name: Install iRODS Externals
+        run:  |
+              wget -qO - https://unstable.irods.org/irods-unstable-signing-key.asc | apt-key add -
+              echo "deb [arch=amd64] https://unstable.irods.org/apt/ $(lsb_release -sc) main" | tee /etc/apt/sources.list.d/renci-irods-unstable.list
+              apt-get update -qq
+              apt-get install -qq \
+                irods-externals-clang13.0.1-0 \
+                irods-externals-cmake3.21.4-0 \
+                irods-externals-fmt-libcxx8.1.1-1 \
+                irods-externals-json3.10.4-0 \
+                irods-externals-jwt-cpp0.6.99.1-0 \
+                irods-externals-nanodbc-libcxx2.13.0-2 \
+                irods-externals-spdlog-libcxx1.9.2-2
+      - name: Configure CMake
+        run:  |
+              export PATH=/opt/irods-externals/cmake3.21.4-0/bin:$PATH
+              mkdir build
+              cmake -GNinja -B build/ -S .
+      - name: Build and Package
+        run:  |
+              export PATH=/opt/irods-externals/cmake3.21.4-0/bin:$PATH
+              cmake --build build/ -t package
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -29,6 +29,7 @@ target_link_libraries(
   "${IRODS_EXTERNALS_FULLPATH_BOOST}/lib/libboost_url.so"
   CURL::libcurl
   jwt-cpp::jwt-cpp
+  "${CMAKE_DL_LIBS}"
 )
 
 target_compile_definitions(


korydraughn:
After thinking about this a little more, making sure the project compiles correctly is more important than the amount of time taken to build it.

Let's remove/disable ccache for now.

korydraughn:
On PRs is good. See irods/irods for reference.

MartinFlores751:
I'll take out cache and put in a separate draft so we won't forget about it

MartinFlores751:
> On PRs is good. See irods/irods for reference.

In that case, only thing to do is remove cache. Already only acts on PRs.

.github/workflows/build-irods-http-api-ubuntu.yml
```diff
@@ -0,0 +1,54 @@
+name: iRODS HTTP API Builder
+on: [pull_request]
+jobs:
+  build:
+    name: Build iRODS HTTP API
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Install Misc
```
korydraughn, line 10: Let's use the same wording as in irods/irods for this.
- `Install Prerequisites`

.github/workflows/build-irods-http-api-ubuntu.yml
```diff
@@ -0,0 +1,54 @@
+name: iRODS HTTP API Builder
```
korydraughn, line 1: Looking at irods/irods, how do you feel about changing this to one of the following?
- `Build-Ubuntu`
- `build-irods-http-api-ubuntu`

Then we could expand this to cover other versions of Ubuntu, and later other platforms such as EL8 and EL9.

korydraughn (commented):
Looks good. Nice and clean.

alanking:
Also, needs an issue number. Seems good overall

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751: I think it would be better without the hyphens. This name is supposed to be a user-facing name. You can see this in the passed checks at the bottom.

Perhaps we can name it so it ends up as `Ubuntu iRODS HTTP API Builder / Build iRODS HTTP API - <Specific Ubuntu Image Here>`?

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751: Avoided that name initially since I wasn't sure what exactly are the prerequisites.

.github/workflows/build-irods-http-api-ubuntu.yml
korydraughn: Sure. I don't think hyphens make it less readable.

Try expanding the builder workflow to support ub22/24, EL8/9 and see how the github actions report it. We want things to be sorted, grouped, and organized.

Aim for consistency across repos.

Take a look at the workflows for irods/irods.

.github/workflows/build-irods-http-api-ubuntu.yml
```diff
@@ -0,0 +1,54 @@
...
+                libssl3 \
+                ninja-build
+      - name: Install iRODS
+        run:  |
+              wget -qO - https://packages.irods.org/irods-signing-key.asc | sudo apt-key add - &&
+              echo "deb [arch=amd64] https://packages.irods.org/apt/ $(lsb_release -sc) main" | sudo tee /etc/apt/sources.list.d/renci-irods.list
+              sudo apt-get update -qq
+              sudo apt-get install -qq \
+                irods-dev=4.3.2-0~jammy \
+                irods-runtime=4.3.2-0~jammy
```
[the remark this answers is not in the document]
MartinFlores751: Did a matrix.

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751: > Try expanding the builder workflow to support ub22/24, EL8/9 and see how the github actions report it. We want things to be sorted, grouped, and organized.

Expanded Ubuntu side.

.github/workflows/build-irods-http-api-ubuntu.yml
```diff
@@ -0,0 +1,62 @@
+name: iRODS HTTP API Builder
+on: [pull_request]
+jobs:
+  build:
+    name: Build - ${{ matrix.os }}, iRODS ${{ matrix.irods-version }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-20.04, ubuntu-22.04, ubuntu-24.04 ]
```
korydraughn, line 9: These entries are not the same as what's used in irods/irods.

Are these container names? irods/irods uses `ubuntu:20.04.`
Is there a difference between `ubuntu-20.04` and `ubuntu:20.04?`
Where is the `container` property? Is that deprecated now?

MartinFlores751: > Are these container names?

The names come from: https://github.com/actions/runner-images

> Where is the `container` property? Is that deprecated now?

`container` is not deprecated.

Using irods/irods as an example:
https://github.com/irods/irods/blob/a6d646c37423f4e8fc25b21d4e15c2025eaed984/.github/workflows/build-irods-ubuntu.yml#L6-L10

Here, I am changing `runs-on`, the github action runner, to then build on the runner.

In irods/irods, the same runner is used, but `container` is set, so a container is created in the runner that then runs the code.

Not sure if this is what you meant, but just in case, the particular name under `matrix` does not matter. `os` could be labeled `images-to-build-and-test-on` for example.

> Is there a difference between ubuntu-20.04 and ubuntu:20.04?

The runner image does have more on it than the base Ubuntu image, to allow for the running of various actions (e.g. ccache action).

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751: > I like the inclusion of iRODS 4.3.x in the entries, but it implies each release of iRODS will have an entry

Yes, each release would have an entry since we state a minimum only.

> unless we come up with a policy for when a dev package should be added/removed.

I'm not sure if I follow, are you asking what if we have a change of requirements or a breaking change?
~If our requirements change, then our tests would match that too, whenever the build container changes, right?~
In terms of iRODS version, isn't this just when our README or build container changes?

> I don't think that's necessary so let's stick with building against the minimum version of the iRODS dev package (i.e. 4.3.2). That aligns with the dev package used to produce the Docker images and it covers backwards compatibility.

We should test the minimum, but I think it would be nice to make sure there are no problems compiling with a more recent version (and eventually test against this too). There's also the concern of accidental forward "departures from interface" or breakages.

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751, line 9: Might have to remove Ubuntu-20.04, or it grows its own file if there's a fix for this:

```console
$ sudo apt-get update -qq
$ sudo apt-get install -qq \
>    apt-transport-https \
>    git \
>    gnupg \
>    lsb-release \
>    wget \
>    g++-11 \
>    gcc-11 \
>    libcurl4-gnutls-dev \
>    libssl-dev \
>    libssl3 \
>    ninja-build

E: Unable to locate package gcc-11
E: Package 'libssl3' has no installation candidate
```

.github/workflows/build-irods-http-api-ubuntu.yml
korydraughn: > I'm not sure if I follow, are you asking what if we have a change of requirements or a breaking change?
~If our requirements change, then our tests would match that too, whenever the build container changes, right?~
In terms of iRODS version, isn't this just when our README or build container changes?

Consider the case where iRODS 4.3.7 is released and 4.3.2 is still the minimum version required for compiling the software.

While it would be nice to build against all versions, it's probably unrealistic to do that in GitHub due to potential restrictions/constraints. We'd need to define rules (a policy) for picking which versions to build against, else people may wonder **why** we build against a subset of the versions instead of all of them.

> We should test the minimum, but I think it would be nice to make sure there are no problems compiling with a more recent version (and eventually test against this too). There's also the concern of accidental forward "departures from interface" or breakages.

I agree. That's something that can be done by the k8s testing environment.

Keep in mind we'll continue to iterate on this. This is a good first step.

.github/workflows/build-irods-http-api-ubuntu.yml
korydraughn: Could try using `uname` or `/etc/os-release` to sniff the OS version and install different packages based on the results. That's if you want to keep it to one file.

Creating a new file is fine too, assuming GitHub won't complain about duplicate workflow names.

korydraughn: Got it.

So why do something different than the irods/irods repo?
Is what you're doing the recommended way of doing things?

_I'm just curious._

MartinFlores751: In a discuss:

We will use the container images instead of the runner images.

Reasoning is to be consistent with other projects.

MartinFlores751: In Discuss:

We will remove Ubuntu 20.04.

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751: In Discuss:

We will implement the other non Ubuntu images in a separate PR.

.github/workflows/build-irods-http-api-ubuntu.yml
trel: ub20 is not EOL until april 2025?

MartinFlores751 (commented):
One question.

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751: Maybe relevant: https://github.com/irods/irods_client_http_api/pull/370#discussion_r1868513336

MartinFlores751: Using container images.

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751: Updated names.

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751: Removed ub20, not resolving, so there can be discussion.

.github/workflows/build-irods-http-api-ubuntu.yml
MartinFlores751: Done.

.github/workflows/build-irods-http-api-ubuntu.yml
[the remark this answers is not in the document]
MartinFlores751: Discuss:

Just use 4.3.3 for u24.

MartinFlores751: Discuss:

Will add ub20, @korydraughn will help out with figuring out the right packages.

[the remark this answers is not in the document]
korydraughn: Yes, use 4.3.3 for Ub24.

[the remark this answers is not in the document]
MartinFlores751: Done.

korydraughn: Use the following:
- `libssl1.1`
- `g++-10`
- `gcc-10`

korydraughn: You'll likely get a linker error for `dlclose`. Not sure why I don't get that error when building directly on the host OS.

Will open an issue for that soon.

MartinFlores751: Should I add the build for 20 here, and then we wait on the issue? Or have 20 separate, like the other images?

korydraughn: Let's add ub20 to this since we're handling the others.
- Add a new commit for the linker error
- Then, add a github action for ub20

MartinFlores751: Rebased onto branch that has fix.

Also added ub20.

Done.

core/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ target_link_libraries(
   "${IRODS_EXTERNALS_FULLPATH_BOOST}/lib/libboost_url.so"
   CURL::libcurl
   jwt-cpp::jwt-cpp
+  ${CMAKE_DL_LIBS}
```
korydraughn, line 32: Please wrap in double quotes.
- `"${CMAKE_DL_LIBS}"`

MartinFlores751: Done.

korydraughn:
Looks like you've done it.

Please squash to two commits.

MartinFlores751:
See if you like the simplified version. If you do, we can keep that, otherwise, can remove the last two commits and squash then.

korydraughn:
All checks passed. Squash it.

MartinFlores751:
Squashed.

MartinFlores751:
Going to change `[_382] Add build to github action` to `[_382] Add Ubuntu builds to github action`

korydraughn (approved):
Pound it.

MartinFlores751:
Pounded.

Issue #375, opened 2024-11-29
Handle missing `aud` claim in access token


ll4strw:
[JSON Web Token (JWT) Profile for OAuth 2.0 Access Tokens](https://datatracker.ietf.org/doc/html/rfc9068#section-2.2) states that access tokens in JWT format must present an `aud` claim. However, [Section 4.1.3](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3) of [[RFC7519](https://datatracker.ietf.org/doc/html/rfc7519)], which defines JWTs, states that the use of `aud` claims is optional.

Regardless the interpretation that one can give to the documents reported above, it is important to properly handle the cases in which an IAM server could omit the `aud` claim in its access tokens. In particular, I propose the rejection of any access tokens missing  the `aud` claim.

NB: keycloak is one of the IAM servers which could issue access tokens without `aud` claim.

MartinFlores751:
> Regardless the interpretation that one can give to the documents reported above, it is important to properly handle the cases in which an IAM server could omit the `aud` claim in its access tokens. In particular, I propose the rejection of any access tokens missing the `aud` claim.

This is an interesting case for this particular endpoint. I agree that we should not accept any access token missing the `aud` claim, as this follows best practice in [OAuth 2.0 Security Best Current Practice (Section 4.10.2)](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#name-audience-restricted-access-). Based on how the introspection endpoint works, though, I'm not sure what is best in this particular case.

I'm not sure if enforcing the return of an `aud` would help for the introspection endpoint. `aud` is optional for this particular endpoint, as the authorization server should handle these checks (see [additional context](#additional-context-for-the-introspection-endpoint) for "gotchas"). Based on the configuration of the server, the introspection endpoint might not return the `aud` member at all, even if the access token has an `aud`.

Enforcing `aud` to be present should allow for us to check the existence of an `aud`, however, if this requires appropriate configuration of the authorization server for this, would the authorization server not handle the check itself? Is there a case where some may not be able to enforce the presence of the `aud` member in the introspection endpoint response?

Perhaps a solution would be an opt-out (or opt-in) to a stricter validation of the introspection endpoint response? My concern is not locking others out of using the HTTP API if they cannot change the authorization server's configuration.

[...]

PR #376, opened 2024-11-29
[#375] Handle missing aud claim in access token

ll4strw:
Reject access tokens without aud claim

### DIFF
--- a/core/src/openid.cpp
+++ b/core/src/openid.cpp
@@ -154,6 +154,12 @@ namespace irods::http::openid
 				logging::warn("{}: Could not find our [client_id] in [aud]. Validation failed.", __func__);
 				return std::nullopt;
 			}
+		// Some IAM servers (e.g. keycloak) could be set up
+		// to exclude `aud' from a bearer token payload 	
+		// If no 'aud' was found in bearer token, do not accept 
+		} else {
+		  logging::warn("{}: Bearer token payload is missing [aud]. Validation failed.", __func__);
+		  return std::nullopt;
 		}
 
 		// The 'iss' provided should match the 'issuer' retrieved from the OpenID Provider's


core/src/openid.cpp
MartinFlores751, line 1: We would need to add the new configuration variable into `core/src/main.cpp`, in the schema:
https://github.com/irods/irods_client_http_api/blob/3824a09215a4b9b7e934178f2a7aa77b9654aa87/core/src/main.cpp#L208-L260

Something like the following should be fine:
```diff
                                "nonstandard_id_token_secret": {{
                                    "type": "string"
                                }},
+                               "strict_introspection_endpoint_aud": {{
+                                   "type": "boolean"
+                               }},
```

After that, we need to add the documentation for the new option in the README:
https://github.com/irods/irods_client_http_api/blob/3824a09215a4b9b7e934178f2a7aa77b9654aa87/README.md?plain=1#L253-L264

MartinFlores751 (changes requested):
The change is very nice and simple. I like it.

Based on the discussions in the issue, we should have this be optional.

I'm suggesting we have a new configuration option `strict_introspection_endpoint_aud` to opt in to the strict check.

The name can change, of course. Open to suggestions.

korydraughn:
@MartinFlores751 You can also make these edits and push to @ll4strw's branch.

korydraughn:
And if you can't do that or feel comfortable doing that, you can create a branch from @ll4strw's, edit it, and then open another PR.

Either approach is fine, but whatever you do, make sure to leave a comment here so the story is clear to everyone.

MartinFlores751:
Gotcha, I'll create a branch and open it in another PR.

core/src/openid.cpp
```diff
@@ -154,6 +154,12 @@ namespace irods::http::openid
 				logging::warn("{}: Could not find our [client_id] in [aud]. Validation failed.", __func__);
 				return std::nullopt;
 			}
+		// Some IAM servers (e.g. keycloak) could be set up
+		// to exclude `aud' from a bearer token payload 	
+		// If no 'aud' was found in bearer token, do not accept 
+		} else {
```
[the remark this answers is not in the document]
MartinFlores751: Handled in #378.

korydraughn:
Once @MartinFlores751 resolves the review comments, we'll give the approval for merging.

core/src/openid.cpp
MartinFlores751: Handled in #378.

core/src/openid.cpp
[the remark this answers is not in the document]
MartinFlores751: This may be resolved.

core/src/openid.cpp
MartinFlores751: This may be resolved.

korydraughn:
Let's hold on merging this until #378 is complete.

korydraughn:
#378 is now complete.

Merging.

korydraughn (approved):

Issue #402, opened 2025-02-03
Server crashes after opening the parallel_write_init for the same lpath


MarekNieslanik:
The server crashes after I try to open the file transfer on the data-object for which I have opened (and not closed) the file transfer already.

Only log I can get from the http_api is 

```
[2025-02-03 10:22:50.114] [P:1] [trace] [T:10] [10.88.0.1] op_parallel_write_init: Opening initial output stream to [/tempZone/home/test_user/aa.data].
```
I would expect some error message in the response.

korydraughn:
What version of the HTTP API are you running?
What version of the iRODS server are you running?

Please post the commands you ran which led to the crash.

MarekNieslanik:
HTTP API version - 0.5.0
iRODS server version - 4.3.2

```
import requests

path = '/tempZone/home/test_platform_user/test.txt'
headers = {
    'Authorization': 'Bearer ' + token
}

data = {
    'op': 'parallel_write_init',
    'lpath': path,
    'stream-count': 5
}

r = requests.post('http://localhost:9000/irods-http-api/0.5.0/data-objects', headers=headers, data=data)
r = requests.post('http://localhost:9000/irods-http-api/0.5.0/data-objects', headers=headers, data=data)
```

korydraughn:
I've reproduced the issue and opened a PR which resolves it.

This will be fixed in the next release.

Issue #399, opened 2025-01-07
Add option to target specific resource or replica number to `parallel_write_init` operation


korydraughn:

korydraughn:
Include a test which shows multiple invocations of `parallel_write_init` with a replica number or resource name for an already open data object do not trigger cause the server to segfault.

This test is an extension of #402.

korydraughn:
Reopening this to add support for replica numbers.

PR #405, opened 2025-02-10
[#402] Server does not segfault on multiple calls to parallel_write_init

korydraughn:
Verified tests trigger the segfault.

Verified the code change resolves the segfault.

### DIFF
--- a/test/test_irods_http_api.py
+++ b/test/test_irods_http_api.py
@@ -2339,6 +2339,75 @@ class test_data_objects_endpoint(unittest.TestCase):
             })
             self.logger.debug(r.content)
 
+    def test_server_does_not_segfault_when_parallel_write_init_is_invoked_multiple_times_on_same_logical_path(self):
+        headers = {'Authorization': 'Bearer ' + self.rodsuser_bearer_token}
+        data_object = f'/{self.zone_name}/home/{self.rodsuser_username}/parallel_write_no_segfaults.txt'
+
+        # Indicates whether the parallel_write_shutdown operation needs to be called.
+        # This is used for clean up when errors occur.
+        invoke_parallel_write_shutdown = False
+
+        try:
+            # Start a parallel-write operation.
+            r = requests.post(self.url_endpoint, headers=headers, data={
+                'op': 'parallel_write_init',
+                'lpath': data_object,
+                'stream-count': 2
+            })
+            self.logger.debug(r.content)
+            self.assertEqual(r.status_code, 200)
+            result = r.json()
+            self.assertEqual(result['irods_response']['status_code'], 0)
+            parallel_write_handle = result['parallel_write_handle']
+
+            # Make sure the parallel_write_shutdown operation is invoked now that we
+            # have a parallel-write handle.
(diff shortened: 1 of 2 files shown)


test/test_irods_http_api.py
```diff
@@ -2339,6 +2339,71 @@ def test_server_correctly_handles_embedded_CRLFs_in_multipart_form_data(self):
...
+            })
+            self.logger.debug(r.content)
+            self.assertEqual(r.status_code, 200)
+            result = r.json()
+            self.assertEqual(result['irods_response']['status_code'], 0)
+            parallel_write_handle = result['parallel_write_handle']
+
+            # Make sure the parallel_write_shutdown operation is invoked now that we
+            # have a parallel-write handle.
+            invoke_parallel_write_shutdown = True
```
alanking, line 2361: I believe this will need to be defined (and set to `False`, I suppose) before the assertions above. The `finally` block will have errors related to `invoke_parallel_write_shutdown` being used before it's defined. I realize the cleanup would not be necessary in that case, but it's just extra noise we don't need and can easily be avoided, IMO.

test/test_irods_http_api.py
```diff
@@ -2339,6 +2339,71 @@ def test_server_correctly_handles_embedded_CRLFs_in_multipart_form_data(self):
...
+        headers = {'Authorization': 'Bearer ' + self.rodsuser_bearer_token}
+        data_object = f'/{self.zone_name}/home/{self.rodsadmin_username}/no_perms_no_segfaults.txt'
+
+        # Show that invoking the parallel_write_init operation fails with an error
+        # if the user does not have the appropriate permissions for opening the data object.
+        r = requests.post(self.url_endpoint, headers=headers, data={
+            'op': 'parallel_write_init',
+            'lpath': data_object,
+            'stream-count': 2
+        })
```
alanking, line 2403: Does it matter that this is attempting to create a data object versus writing to an existing one? Just wondering whether there ought to be a test for each scenario. "No" is a perfectly fine answer.

alanking (commented):
Good fix. Just some comments on the tests

test/test_irods_http_api.py
korydraughn: Agreed.

Copy-paste got me. Will update.

test/test_irods_http_api.py
korydraughn: It doesn't matter because everything uses dstream.

I'll add a test for that later. As of right now, the parallel_write_init operation doesn't support targeting a specific replica. We have an issue to address that though. Can handle all of this along with that.

korydraughn: This conversation is now linked to #399.

test/test_irods_http_api.py
korydraughn: Done.

alanking (commented):
Looks good to me. Squash em

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Pounded.

Issue #395, opened 2024-12-17
Add a docker compose file for testing environment


MartinFlores751:
Currently, the HTTP API testing environment requires a degree of manual setup.

It would be nice to have the HTTP API and the Keycloak image for testing be set up automatically alongside iRODS.

korydraughn:
The following conversation needs to be considered when handling this issue: https://github.com/irods/irods_client_http_api/pull/424#discussion_r2211627872

PR #424, opened 2025-07-16
Implement improvements for write operation / Fix build issues with iRODS 5 (main)

korydraughn:
All non-OIDC tests pass except for ones which have already been reported as needing tweaks.

The surprising thing is that only one test failed for me. The failure isn't related to the changes in this PR.
```
======================================================================                                             
FAIL: test_report_operation (__main__.test_zones_endpoint)                                                         
  File "/mnt/hdd_3tb/github/irods_client_http_api/test/test_irods_http_api.py", line 4149, in test_report_operation
    self.assertIn('schema_version', zone_report)                                                                   
AssertionError: 'schema_version' not found in {'zones': <snip>
```

Not a blocker for this PR. Will handle reported failures in a separate PR.

The new tests are clones of existing tests with some tweaks to trigger the new code paths.


.clang-tidy
```diff
@@ -71,6 +71,7 @@ WarningsAsErrors:
     readability-*,
     -readability-avoid-const-params-in-decls,
     -readability-function-cognitive-complexity,
+    -readability-identifier-length,
```
korydraughn, line 74: Propagate this change to the other repos.

endpoints/data_objects/src/main.cpp
```diff
@@ -2373,3 +2554,243 @@ namespace
...
+                            req.version(),
+                            req.keep_alive(),
+                            std::move(conn),
+                            std::move(tp),
+                            std::move(out),
+                            out_ptr,
+                            std::move(mark_pw_stream_as_usable),
+                            max_number_of_bytes_per_write,
+                            is_parallel_write)->start();
+                    // clang-format off
```
korydraughn, line 2761: Fix indentation.

endpoints/data_objects/src/main.cpp
```diff
@@ -2373,3 +2554,243 @@ namespace
...
+			// clang-format off
+                    res.body() = json{
+                            {"irods_response", {
+                                    {"status_code", e.code()},
+                                    {"status_message", e.client_display_what()}
+                            }}
+                    }.dump();
+			// clang-format on
+			res.prepare_payload();
+			_sess_ptr->send(std::move(res));
```
korydraughn, line 2787: Fix indentation.

endpoints/data_objects/src/main.cpp

endpoints/data_objects/src/main.cpp

test/test_irods_http_api.py
```diff
@@ -1825,6 +1825,332 @@ def test_non_parallel_writes_with_data_exceeding_internal_write_threshold(self):
...
+            self.logger.debug(r.content)
+            self.assertEqual(r.status_code, 200)
+            self.assertEqual(r.json()['irods_response']['status_code'], 0)
+
+    def test_parallel_writes_using_header_based_form_with_data_exceeding_internal_write_threshold(self):
+        # This test assumes the HTTP API is configured to use a value smaller than
+        # 96kb for "/irods_client/max_number_of_bytes_per_write_operation". This is
+        # important because values greater than or equal to 96kb will result in the
+        # background thread pool only being used one time. This test is designed to
+        # force the HTTP API to schedule multiple writes across the thread pool.
```
alanking, line 1987: Is it possible to check the configuration and `raise` an `EnvironmentError` if this is the not the case?

test/test_irods_http_api.py
```diff
@@ -1825,6 +1825,332 @@ def test_non_parallel_writes_with_data_exceeding_internal_write_threshold(self):
...
+            # Remove the data object.
+            r = requests.post(self.url_endpoint, headers=headers, data={
+                'op': 'remove',
+                'lpath': data_object,
+                'catalog-only': 0,
+                'no-trash': 1
+            })
+            self.logger.debug(r.content)
+            self.assertEqual(r.status_code, 200)
+            self.assertEqual(r.json()['irods_response']['status_code'], 0)
```
alanking, line 1980: I noticed that `test_writing_to_data_object_using_header_based_form` and the other new tests did not make assertions after removing its data object. Not saying this is right or wrong, just pointing out that this one is different.

test/test_irods_http_api.py
```diff
@@ -1825,6 +1825,332 @@ def test_non_parallel_writes_with_data_exceeding_internal_write_threshold(self):
...
+        headers = {'Authorization': f'Bearer {self.rodsuser_bearer_token}'}
+        data_object = f'/{self.zone_name}/home/{self.rodsuser_username}/parallel_writes_header_based_1.txt'
+
+        # Generate 96kb string.
+        # This is the data we'll upload to the HTTP API.
+        data96kb = ('A' * 32768) + ('B' * 32768) + ('C' * 32768)
+
+        # Calculate a checksum for data.
+        # The checksum will be used to confirm the data was uploaded correctly.
+        checksum = base64.b64encode(hashlib.sha256(data96kb.encode('utf-8')).digest()).decode('utf-8')
```
alanking, line 2005: Not that we're super concerned about speed here, but since we already know the contents of the data object ahead of time, we can just hard-code the checksum here if we wanted to save some cycles.

test/test_irods_http_api.py
```diff
@@ -1825,6 +1825,332 @@ def test_non_parallel_writes_with_data_exceeding_internal_write_threshold(self):
...
+            self.assertEqual(result['irods_response']['status_code'], 0)
+            parallel_write_handle = result['parallel_write_handle']
+
+            # Make sure the parallel_write_shutdown operation is invoked now that we
+            # have a parallel-write handle.
+            invoke_parallel_write_shutdown = True
+
+            # Write to the data object using the parallel write handle.
+            futures = []
+            with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
```
alanking, line 1923: Should this match `stream-count`, or does it just happen to be the same value?

test/test_irods_http_api.py
```diff
@@ -1825,6 +1825,332 @@ def test_non_parallel_writes_with_data_exceeding_internal_write_threshold(self):
             })
             self.logger.debug(r.content)
 
+    def test_writing_to_data_object_using_header_based_form(self):
+        headers = {'Authorization': f'Bearer {self.rodsuser_bearer_token}'}
+        data_object = f'/{self.zone_name}/home/{self.rodsuser_username}/header_based_write_op1.txt'
+
+        # Generate 512kb of random bytes.
+        # This is the data we'll upload to the HTTP API.
+        data512kb = os.urandom(512 * 1024)
```
alanking, line 1834: Any reason this is using random and the other tests use pre-determined contents? Just curious.

test/test_irods_http_api.py
```diff
@@ -1825,6 +1825,332 @@ def test_non_parallel_writes_with_data_exceeding_internal_write_threshold(self):
...
+
+            # Read the contents of the data object and show it contains exactly what we expect.
+            r = requests.get(self.url_endpoint, headers=headers, params={
+                'op': 'read',
+                'lpath': data_object,
+                'count': 30
+            })
+            self.logger.debug(r.content)
+            self.assertEqual(r.status_code, 200)
+            self.assertEqual(r.content.decode('utf-8'), 'A' * 10 + 'B' * 10 + 'C' * 10)
```
alanking, line 1959: Why are we not using the checksum in this test? Just curious.

API.md
```diff
@@ -870,10 +870,14 @@ If there was an error, expect an HTTP status code in either the 4XX or 5XX range
 
 Writes bytes to a data object.
 
+This operation supports two forms of sending data to the HTTP API server. Prefer **form 2**.
```
alanking, line 873: Should we consider Form 1 deprecated? Or is there still a use case for it?

Also, a slight tweak to the wording... Feel free to ignore.
```suggestion
This operation supports two forms of sending data to the HTTP API server. **Form 2** is the recommended form.
```

API.md
```diff
@@ -888,9 +892,33 @@ curl http://localhost:<port>/irods-http-api/<version>/data-objects \
     [-F,--data-urlencode] 'stream-index=<integer>' # The stream to use when writing in parallel. Optional.
 ```
 
-This is the full set of parameters supported by the operation.
+This form is the original implementation. It sends all information via the HTTP request body. The HTTP API server will buffer the full request before processing it.
+
+When sending large amounts of data or writing in parallel, prefer multipart/form-data (`-F`) over application/x-www-form-urlencoded (`--data-urlencode`) as the Content-Type.
```
alanking, line 897:
```suggestion
When sending large amounts of data or writing in parallel, multipart/form-data (`-F`) is recommended over application/x-www-form-urlencoded (`--data-urlencode`) as the Content-Type.
```

core/src/session.cpp
```diff
@@ -21,6 +21,13 @@
 #  include <fstream>
 #endif
 
+namespace irods::http::endpoint_operation
+{
+	// This forward declaration is needed for streaming write operations.
+	// See endpoints/data_objects/src/main.cp for the implementation.
+	auto op_write_streaming(irods::http::session_pointer_type _sess_ptr) -> void;
+} // namespace irods::http::endpoint_operation
```
alanking, line 25: I am curious as to why the implementation ended up in main since it only seems to be used in this file. Not saying it's wrong, just wondering whether it was brought into consideration.

alanking (commented):
Seems good. Nothing in the actual implementation stuck out as being wrong. Most of my feedback is questions

test/test_irods_http_api.py
korydraughn: Yes, that's possible if we want to make it a rule that the config is always local to the test.

API.md
korydraughn: Form 1 is still valid. There might be a use-case for it, but it's difficult to assert that right now. I think we leave it in until we have more information from the community.

Will update wording.

core/src/session.cpp
korydraughn: The implementation lives in data_objects/src/main.cpp because it's part of the data objects write operation family. The headers for the implementation were already available too. I also wanted to keep session.cpp focused on entry-related logic.

So, it's primarily about keeping related things in the same place.

test/test_irods_http_api.py
korydraughn: Yes, it should match `stream-count`. Will update that.

test/test_irods_http_api.py
korydraughn: This may be from a time when the checksum operation wasn't quite there yet. Or, perhaps the test which this was derived from was attempting to prove the "read" operation worked as intended.

test/test_irods_http_api.py
korydraughn: These assertions shouldn't be there. The reason there's a difference is due to the age of tests. I believe the tests which do not contain these assertions are newer than the one which I derived this test from.

test/test_irods_http_api.py
korydraughn: Very true. There are likely other instances of this in the test suite. I want to keep one around so it's easy to find. You never know when you might need to compute a checksum in python.

Will confirm there are other instances of this before updating.

test/test_irods_http_api.py
korydraughn: I believe we use random to make sure the HTTP API doesn't fall over on special characters and byte sequences.

While we don't know which characters can/will cause a problem, this test might reveal such a case.

trel: generally we want our tests to be reproducible.    if this fails one time... is that actionable if it's random?

core/src/session.cpp
```diff
@@ -21,6 +21,13 @@
 #  include <fstream>
 #endif
 
+namespace irods::http::endpoint_operation
+{
+	// This forward declaration is needed for streaming write operations.
+	// See endpoints/data_objects/src/main.cp for the implementation.
```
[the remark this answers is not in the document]
korydraughn: Done.

API.md

API.md
korydraughn: Wording updated.

@alanking Leaving open in case you have more questions about form 1.

test/test_irods_http_api.py

API.md
alanking: Makes sense to me. No further questions.

test/test_irods_http_api.py
alanking: Okay. I feel like programmatically enforcing our test prerequisites could be helpful. Then again, if the configuration were to change and this test fails, the comment is here to explain why the test may have failed. Up to you.

test/test_irods_http_api.py
alanking: > I believe we use random to make sure the HTTP API doesn't fall over on special characters and byte sequences.

If that's the case, should we not be doing this for all the tests? I just wonder why this test specifically is getting the fuzzing treatment. I guess that's why it stood out to me in the first place.

test/test_irods_http_api.py
korydraughn: After some discussion, we've decided to assume that the tests are always run in the same environment as the HTTP API. That is, the test suite should always be able to read the config file of the HTTP API.

The new tests which this rule applies to will be updated to raise an `EnvironmentError` if the rule is violated.

test/test_irods_http_api.py
korydraughn: Removed assertions.

test/test_irods_http_api.py
korydraughn: Hmm, right after I started working to update the test, I realized that the location of the config file can be anywhere on the computer.

I'm going to hold on enforcing the prereqs and handle it as part of #395.

korydraughn: Left a comment in #395 which points to this conversation.

test/test_irods_http_api.py
korydraughn: Done.

test/test_irods_http_api.py
korydraughn: We can, but I don't know that we need to. Happy to hear opinions.

Perhaps there's a well-defined list of byte sequences that HTTP servers are known to struggle with. If we had such a list, we could test against it and limit use of `os.urandom` to when we truly need it.

korydraughn: > To me, this kind of thing just introduces additional things that the test is trying to do, where, as a general (and, for my part, hypocritical) matter, I prefer more focus in what behavior the test is trying to verify.

I totally agree that tests should focus on one thing as much as possible.

> I suppose it doesn't make much sense to have a separate test for such things if it will be performing the same actions. So maybe this is exactly the right place for it.

Right. This kind of situation also comes up when thinking about 

> I think that what we have is fine. I was just curious as to why it was in this test and not the others, mainly.

The change was me seeing a relatively free opportunity to increase the chances of revealing issues stemming from binary data.

There wasn't any deep thought behind that one change.

alanking (approved):
Seems good to me. # it.

Issue #431, opened 2025-07-24
iRODS 5: Add support for modifying the atime of a replica


korydraughn:
## Enhancement

This is to allow admins to modify the atime of a replica via the modify_replica operation.

Initial implementation is here: https://github.com/irods/irods_client_http_api/commit/217c3e35658329e763880ee3ffd69ce9b0a9c7e1

PR #439, opened 2025-07-26
[#431] Add support for modifying the atime of a replica

korydraughn:
All non-OIDC tests pass.

A HTTP status code of 400 (Bad Request) is returned to the client if the atime is specified in the request and the HTTP API is connected to an iRODS server earlier than 5.0.0. This decision guarantees the client will have a consistent experience around the use of the atime property.

This PR does make it more difficult to test 4.3 servers due to the atime property being used in a test. #440 will help with fixing that issue.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1181,7 +1181,8 @@ If there was an error, expect an HTTP status code in either the 4XX or 5XX range
 
 Modifies properties of a single replica.
 
-**WARNING:** This operation requires rodsadmin level privileges and should only be used when there isn't a safer option. Misuse can lead to catalog inconsistencies and unexpected behavior.
+> [!WARNING]
+> This operation requires rodsadmin level privileges and should only be used when there isn't a safer option. Misuse can lead to catalog inconsistencies and unexpected behavior.
 
 #### Request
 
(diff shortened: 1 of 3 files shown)


endpoints/data_objects/src/main.cpp
```diff
@@ -2527,8 +2535,17 @@ namespace
...
+					constexpr std::string_view atime_prop = "new-data-access-time";
 					for (auto&& [external_pname, internal_pname] : properties) {
 						if (const auto iter = _args.find(external_pname); iter != std::end(_args)) {
+							// TODO: Consider removing this.
+							// TODO: Check if older servers return an error on unknown properties.
+							// TODO: If this is kept, don't forget to handle the other locations.
+							if (!connected_to_irods5_or_later && external_pname == atime_prop) {
+								logging::error(*_sess_ptr, "{}: iRODS server does not support access time.", fn);
+								return _sess_ptr->send(irods::http::fail(res, http::status::bad_request));
+							}
```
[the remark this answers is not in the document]

test/test_irods_http_api.py
```diff
@@ -2354,12 +2354,18 @@ def test_modifying_replica_properties(self):
             self.assertEqual(result['rows'][0][1], '0')
 
             # Change the replica's status and data size using the modify_replica operation.
```
trel, line 2356:
```suggestion
            # Change the replica's status, data size, and access time using the modify_replica operation.
```

API.md
```diff
@@ -1207,11 +1208,14 @@ curl http://localhost:<port>/irods-http-api/<version>/data-objects \
     --data-urlencode 'new-data-size=<integer>' \
     --data-urlencode 'new-data-status=<string>' \
     --data-urlencode 'new-data-type-name=<string>' \
-    --data-urlencode 'new-data-version=<string>'
+    --data-urlencode 'new-data-version=<string>' \
+    --data-urlencode 'new-data-access-time=<string>' # Must be 11 bytes in length and padded with leading zeros.
```
trel, line 1212: should we consider NOT requiring the clients to know about our non-standard storage of 11 bytes padded with leading zeroes?   can we just take in a normal number-as-string and do the padding ourselves in the HTTP API before sending to iRODS?

is that more confusing since we return the raw database value when someone asks for times from the database?

korydraughn: Yes, we can do that. We need to determine if there are any downsides to that.

Consider the case where a user receives a raw timestamp from the catalog and attempts to pass it to the API. The user would have to adjust the timestamp. The HTTP API is now also working with different formats for timestamps (presentation vs input).

trel: right - so perhaps we just make the HTTP API accept 'both'?       we can detect if it's not 11-bytes, and pad if necessary?    maybe that's too much magic...

alanking (commented):
Nothing seemed out of place here to me. I say squash

alanking (approved):
\# it

Issue #443, opened 2025-07-31
Automatically use the default TLS directory, have `tls_certificates_directory` be optional


MartinFlores751:
Use `.set_default_verify_paths()` in `transport.cpp` to use the default TLS directory for the install.

korydraughn:
@MartinFlores751 Please close if complete.

PR #444, opened 2025-07-31
[#443] Set default TLS cert directory (main)

MartinFlores751:
Issue quick link: #443 

This PR attempts to automatically set the TLS certificate directory to the default for the system. Additionally, `tls_certificates_directory` becomes an optional custom certificate directory which may point to a non-default location for additional TLS certificates.

---

Before pulling out of draft:
- [ ] Format code.
- [ ] Test setting `tls_certificates_directory` (the default certs should still be included).

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,8 +290,8 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g. 
 
                 // The path to the TLS certificates directory used to establish
                 // secure connections between the HTTP API and the OpenID provider.
-                // Typically, this should point to a standard certificate directory
-                // such as "/etc/ssl/cert".
+                // This should not be set unless your TLS certificates directory is
+                // not a standard certificate directory such as "/etc/ssl/cert".
                 "tls_certificates_directory": "/path/to/certs"
             }
         },
--- a/core/src/main.cpp
+++ b/core/src/main.cpp
@@ -269,7 +269,6 @@ constexpr auto default_jsonschema() -> std::string_view
                                 "provider_url",
                                 "client_id",
                                 "require_aud_member_from_introspection_endpoint",
-                                "tls_certificates_directory",
                                 "user_mapping",
                                 "client_secret"
                             ]
--- a/core/src/transport.cpp
+++ b/core/src/transport.cpp
@@ -134,11 +134,13 @@ namespace irods::http
 	{
 		boost::asio::ssl::context ctx{boost::asio::ssl::context::tlsv12_client};
 		const auto config{irods::http::globals::oidc_configuration()};
+		ctx.set_default_verify_paths();
 
-		const std::string& cert_path{config.at("tls_certificates_directory").get_ref<const std::string&>()};
-		ctx.add_verify_path(cert_path);
+		if (const auto cert_path{config.find("tls_certificates_directory")}; cert_path != std::end(config)) {
+		  ctx.add_verify_path(cert_path->get_ref<const std::string&>());
+		}
+
 		ctx.set_verify_mode(boost::asio::ssl::verify_peer);
-
 		return ctx;
 	}
 


README.md
```diff
@@ -290,8 +290,8 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
 
                 // The path to the TLS certificates directory used to establish
                 // secure connections between the HTTP API and the OpenID provider.
-                // Typically, this should point to a standard certificate directory
-                // such as "/etc/ssl/cert".
+                // This should not be set unless your TLS certificates directory is
+                // not a standard certificate directory such as "/etc/ssl/cert".
```
korydraughn, line 294: Consider rewording this. It is a little hard to comprehend due to multiple uses of the word "not".

Try stating when this property SHOULD be used rather than when it shouldn't.

core/src/transport.cpp
```diff
@@ -134,11 +134,13 @@ namespace irods::http
 	{
 		boost::asio::ssl::context ctx{boost::asio::ssl::context::tlsv12_client};
 		const auto config{irods::http::globals::oidc_configuration()};
+		ctx.set_default_verify_paths();
 
-		const std::string& cert_path{config.at("tls_certificates_directory").get_ref<const std::string&>()};
-		ctx.add_verify_path(cert_path);
-		ctx.set_verify_mode(boost::asio::ssl::verify_peer);
+		if (const auto cert_path{config.find("tls_certificates_directory")}; cert_path != std::end(config)) {
+			ctx.add_verify_path(cert_path->get_ref<const std::string&>());
```
korydraughn, line 140: Does this ADD to the set of paths to search or override it?

trel: i hope add.    add is what we want i think.

alanking: Seems pretty easy to test, in theory. Put certificates in the default location, specify an empty `tls_certificates_directory`, see if it works?

korydraughn: Adding to what @alanking said - can also move a self-signed cert between the two locations to see how the HTTP API behaves.

korydraughn: > We could test a self-signed cert, but I would rather move all the default certs to a separate directory, leaving the default cert directory empty, and point `tls_certificates_directory` to the new directory to see that everything still works fine.

Are you talking about moving the system certs from their normal location?

core/src/transport.cpp
```diff
@@ -134,11 +134,13 @@ namespace irods::http
...
+		}
```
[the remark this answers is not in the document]
MartinFlores751: Did a quick test, will push up shortly.

---

Moving the default certs as I did before (see https://github.com/irods/irods_client_http_api/pull/444#discussion_r2252672679).

Setting `tls_certificates_directory` to `[ "/", "/bleh", "/workspace" ]` fails as expected
Setting `tls_certificates_directory` to `[ "/", "/bleh/certs", "/workspace" ]` works as expected.

core/src/main.cpp
```diff
@@ -501,7 +503,7 @@ auto print_configuration_template() -> void
                     "configuration": {{
                     }}
                 }},
-                "tls_certificates_directory": "<string>"
+                "tls_certificates_directories": [ "<string>" ]
```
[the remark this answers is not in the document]
MartinFlores751: Done.

README.md
MartinFlores751: > Please update to be closer to the wording @trel provided.

Will update depending on what is decided here: https://github.com/irods/irods_client_http_api/pull/444#discussion_r2254286786

README.md
```diff
@@ -288,11 +288,11 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
                     }
                 },
 
-                // The path to the TLS certificates directory used to establish
+                // A list of paths to the TLS certificates directories used to establish
                 // secure connections between the HTTP API and the OpenID provider.
                 // This should not be set unless your TLS certificates directory is
                 // not a standard certificate directory such as "/etc/ssl/cert".
```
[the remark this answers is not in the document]
MartinFlores751: > > I'm kind of torn on it. I think it would be helpful to know what the Normal and Boring / expected behavior is.
> 
> @MartinFlores751 See how other popular server applications handle this in their config files. For instance, nginx, apache httpd, postgresql, etc.

I took a look at other OIDC clients (because we still are a client in the OIDC world, just protecting a resource), and from what I can see not every client supports setting custom CA directories. Of those that do, none seem to include the default cert directory automatically if the cert is set in a configuration option.

In the case for an Apache module, it decides to use the defaults unless a user-provided certificate directory is given:
https://github.com/OpenIDC/mod_auth_openidc/blob/72304a45eb3d3b5caa42ec5ee2e0cc81ad5b1f38/auth_openidc.conf#L187-L189

And in the case of an erlang library, it seems to allow you to configure whatever certs you want by specifying the path to the certs or passing the cert information directly (see oidcc erlang library).

Thinking a bit more about it, it's fine to have the user type out the default cert directories as well if they are adding a custom one. It's similar to the original behavior prior to this PR, and it gives the most choice in what is loaded.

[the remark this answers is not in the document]
alanking: So it sounds like the behavior described by @korydraughn is the way to go? 

> Defining this option should result in the default path no longer being included automatically. This gives admins total control over the search directories.

README.md
MartinFlores751: Updated the wording.

README.md
```diff
@@ -288,11 +288,12 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
...
-                // The path to the TLS certificates directory used to establish
+                // A list of paths to the TLS certificates directories used to establish
                 // secure connections between the HTTP API and the OpenID provider.
-                // Typically, this should point to a standard certificate directory
-                // such as "/etc/ssl/cert".
-                "tls_certificates_directory": "/path/to/certs"
+                // Use this setting to point to a non-standard certificate directory.
+                // When set, this will not include the standard certificate directory
+                // by default.
+                "tls_certificates_directories": ["/path/to/certs"]
```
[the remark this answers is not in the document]

korydraughn:
Consider rephrasing the commit message body. I'm finding it a little difficult to reason about.

MartinFlores751:
Rephrased.

korydraughn (approved):
Pound it.

alanking (approved):

Issue #448, opened 2025-08-01
Make OIDC Access Token validation method a configuration option


MartinFlores751:
Currently, the HTTP API defaults to performing local validation and then falling through to using the introspection endpoint if local validation fails. Having both validation methods enabled provides no benefit, since they both perform the same job through different methods. We should allow for one validation method to run at a time and have the HTTP API administrator choose the appropriate method based on their OpenID Provider.

An idea for the configuration would be an enum allowing `local_validation` or `introspection`. This would allow only one method to run at a time:
```
{
  "access_token_validation_method": "local_validation"
}
```

MartinFlores751:
> two questions...
> 
>     * does this introduce a tripping hazard by requiring the administrator to know more about the openid provider?   or is it better because it reduces the 'magic' being performed today... aka 'try one, then try the next, and not sure which might succeed'?

It's better by removing 'magic'. The most the admin needs to know is if the OpenID provider they're using provides an introspection endpoint. If they do, then you can use it. Otherwise, you can only use `local_validation`. This would also show up in the logs if the introspection endpoint does not exist.

>     * this would be a new required configuration option?   or, if not defined...`local_validation` would be the default?

I think a new required option (e.g. `access_token_validation_method`) that defaults to `local_validaiton` when you dump the sample configuration is best.

trel:
Both sound good to me.  No magic and a default when dumped.

MartinFlores751:
> [@MartinFlores751](https://github.com/MartinFlores751) Please close if complete. Consider closing [#380](https://github.com/irods/irods_client_http_api/issues/380) too if it is a duplicate or is no longer necessary. Add the "resolved / invalid" tag to it and leave a comment stating which PR resolved it. Don't forget to move that other issue into the 0.6.0 milestone depending on if you close it.

I'll close the issue and mark it as a duplicate.

PR #450, opened 2025-08-05
[#448] Make access token validation method selectable (main)

MartinFlores751:
Issue quick link: #448

This PR introduces a new schema item to select the desired access token validation method. The selected method will be the only method that runs until the configuration is changed and the server is restarted.

If you have suggestions for the schema item name and names for the enums, they would be appreciated.

---

Before getting out of draft:

- [x] Ensure only one access token validation method is executed at a time, and that the method executed is the one selected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,15 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g. 
                 // times out, requiring another attempt at authentication.
                 "state_timeout_in_seconds": 600,
 
+                // The desired method with which to validate OIDC access tokens.
+                // The two methods supported are "local_validation" and
+                // "introspection".
+                //
+                // The following values are supported:
+                // - local_validation: Validate JWT access tokens according to RFC 9086.
+                // - introspection: Validate access tokens according to RFC 7662.
+                "access_token_validation_method": "local_validation",
+
                 // Defines relevant information related to the User Mapping plugin system.
                 // Allows for the selection and configuration of the plugin.
                 "user_mapping": {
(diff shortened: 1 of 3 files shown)


core/src/main.cpp
```diff
@@ -241,6 +241,9 @@ constexpr auto default_jsonschema() -> std::string_view
                                 "access_token_secret": {
                                     "type": "string"
                                 },
+                                "access_token_validation_method": {
```
alanking, line 244: Should we to add this to the list of "required" keys? The implementation will throw an exception if it is not present.

MartinFlores751: Done.

I see the following failure now with the change:
```
...
Configuration failed validation.
[
    {
        "valid": false, 
        "evaluationPath": "/properties/http_server/properties/authentication/properties/openid_connect/required", 
        "schemaLocation": "https://schemas.irods.org/irods-http-api/config.json#/properties/http_server/properties/authentication/properties/openid_connect/required", 
        "instanceLocation": "/http_server/authentication/openid_connect", 
        "error": "Required property 'access_token_validation_method' not found."
    }
]
```

README.md
```diff
@@ -272,6 +272,15 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
                 // times out, requiring another attempt at authentication.
                 "state_timeout_in_seconds": 600,
 
+                // The desired method with which to validate OIDC access tokens.
+                // The two methods supported are "local_validation" and
+                // "introspection".
+                //
+                // The following values are supported:
+                // - local_validation: Validate JWT access tokens according to RFC 9086
+                // - introspection: Validate access tokens according to RFC 7662
```
korydraughn, line 281: Missing period at the end of each item. Adding the periods make it align with documentation for other config options.

MartinFlores751:
Created the issue for tests and squashed.

MartinFlores751:
Added a message to the commit.

PR #454, opened 2025-08-15
Fix JSON schema config / Improve support for targeting specific replicas (main)

korydraughn:
The commits which mention `squashing w/ ... op` are ones which I'm still thinking about.

The iRODS server checks for bad combinations of options, so those commits are me leaning towards letting the HTTP API set them and the user receiving an error. No need to duplicate work the iRODS server already covers.

Haven't run the full test suite yet. The new tests passed before the squash commits, so I don't expect them to fail with the squash commits.

Still need to update the API docs too.


endpoints/data_objects/src/main.cpp
```diff
@@ -850,7 +850,53 @@ namespace
...
+							logging::error(
+								*_sess_ptr, "{}: [resource] and [replica-number] parameters are incompatible.", fn);
+							res.result(http::status::bad_request);
+							res.prepare_payload();
+							return _sess_ptr->send(std::move(res));
+						}
+
+						int value = -1;
+						try {
+							value = std::stoi(iter->second);
```
alanking, line 885: Just checking: We're leaning on the server to catch invalid replica numbers (i.e. negative values)?

endpoints/data_objects/src/main.cpp
```diff
@@ -1657,8 +1657,25 @@ namespace
 					}
 
 					if (const auto iter = _args.find("src-resource"); iter != std::end(_args)) {
+						if (_args.find("src-replica-number") != std::end(_args)) {
+							logging::error(
+								*_sess_ptr,
+								"{}: [src-resource] and [src-replica-number] parameters are incompatible.",
+								fn);
+							return _sess_ptr->send(irods::http::fail(http::status::bad_request));
+						}
```
alanking, line 1673: The DataObjRepl API enforces that `RESC_NAME_KW` and `REPL_NUM_KW` not be used together, if that's interesting here. Would need to make the `else if` below just an `if` if we wanted to lean on that. What you have here is probably better, but just thought I would mention it.

I moved to the "squash" commits and see that you considered this case as well. :) I guess we can either use this thread to discuss or just resolve this. Up to you.

test/test_irods_http_api.py
alanking, line 1: Would like to see some tests around invalid inputs as well (mostly just the ones enforced by this client - don't have to rehash testing the DataObjRepl API).

alanking (commented):
Looks good overall. Had a couple suggestions.

endpoints/data_objects/src/main.cpp
korydraughn: After some discussion, we've agreed on the following:
- The HTTP API is a passthrough implementation
- The HTTP API will NOT check for incompatible parameters
- The HTTP API can check type information, but should NOT duplicate iRODS server functionality

These rules should be documented in the repo and help keep the HTTP API future-proof with changes in iRODS server behavior.

endpoints/data_objects/src/main.cpp
```diff
@@ -850,7 +850,55 @@ namespace
...
+					if (auto iter = _args.find("resource"); iter != std::end(_args)) {
+#if 0
+						if (_args.find("replica-number") != std::end(_args)) {
+							logging::error(
+								*_sess_ptr, "{}: [resource] and [replica-number] parameters are incompatible.", fn);
+							res.result(http::status::bad_request);
+							res.prepare_payload();
+							return _sess_ptr->send(std::move(res));
+						}
+#endif
```
korydraughn, line 864: Remove this. We've decided that the HTTP API should behave as a passthrough.

endpoints/data_objects/src/main.cpp
```diff
@@ -850,7 +850,55 @@ namespace
...
+					else if (iter = _args.find("replica-number"); iter != std::end(_args)) {
+#if 0
+						if (_args.find("resource") != std::end(_args)) {
+							logging::error(
+								*_sess_ptr, "{}: [resource] and [replica-number] parameters are incompatible.", fn);
+							res.result(http::status::bad_request);
+							res.prepare_payload();
+							return _sess_ptr->send(std::move(res));
+						}
+#endif
```
korydraughn, line 882: Remove this. We've decided that the HTTP API should behave as a passthrough.

endpoints/data_objects/src/main.cpp
```diff
@@ -895,7 +943,55 @@ namespace
...
+				if (auto iter = _args.find("resource"); iter != std::end(_args)) {
+#if 0
+					if (_args.find("replica-number") != std::end(_args)) {
+						logging::error(
+							*_sess_ptr, "{}: [resource] and [replica-number] parameters are incompatible.", fn);
+						res.result(http::status::bad_request);
+						res.prepare_payload();
+						return _sess_ptr->send(std::move(res));
+					}
+#endif
```
korydraughn, line 957: Remove this. We've decided that the HTTP API should behave as a passthrough.

endpoints/data_objects/src/main.cpp
```diff
@@ -895,7 +943,55 @@ namespace
...
+				else if (iter = _args.find("replica-number"); iter != std::end(_args)) {
+#if 0
+					if (_args.find("resource") != std::end(_args)) {
+						logging::error(
+							*_sess_ptr, "{}: [resource] and [replica-number] parameters are incompatible.", fn);
+						res.result(http::status::bad_request);
+						res.prepare_payload();
+						return _sess_ptr->send(std::move(res));
+					}
+#endif
```
korydraughn, line 975: Remove this. We've decided that the HTTP API should behave as a passthrough.

endpoints/data_objects/src/main.cpp
korydraughn: > Just checking: We're leaning on the server to catch invalid replica numbers (i.e. negative values)?

Yes. We're leaning on the iRODS server to catch invalid replica numbers.

> After some discussion, we've agreed on the following:
>
> - The HTTP API is a passthrough implementation
> - The HTTP API will NOT check for incompatible parameters
> - The HTTP API can check type information, but should NOT duplicate iRODS server functionality
>
> These rules should be documented in the repo and help keep the HTTP API future-proof with changes in iRODS server behavior.

Created #456.

endpoints/data_objects/src/main.cpp
korydraughn: We've discussed this. See https://github.com/irods/irods_client_http_api/pull/454#discussion_r2279738616.

endpoints/data_objects/src/main.cpp

endpoints/data_objects/src/main.cpp

endpoints/data_objects/src/main.cpp
```diff
@@ -1086,9 +1182,43 @@ namespace
...
 					if (const auto iter = _args.find("resource"); iter != std::end(_args)) {
+#if 0
+						if (_args.find("replica-number") != std::end(_args)) {
+							logging::error(
+								*_sess_ptr, "{}: [resource] and [replica-number] parameters are incompatible.", fn);
+							res.result(http::status::bad_request);
+							res.prepare_payload();
+							return _sess_ptr->send(std::move(res));
+						}
+#endif
```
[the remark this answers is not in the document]

endpoints/data_objects/src/main.cpp
```diff
@@ -1533,9 +1663,31 @@ namespace
...
+					if (const auto iter = _args.find("src-replica-number"); iter != std::end(_args)) {
+#if 0
+						if (_args.find("src-resource") != std::end(_args)) {
+							logging::error(
+								*_sess_ptr,
+								"{}: [src-resource] and [src-replica-number] parameters are incompatible.",
+								fn);
+							return _sess_ptr->send(irods::http::fail(http::status::bad_request));
+						}
+#endif
```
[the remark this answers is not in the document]

endpoints/data_objects/src/main.cpp
```diff
@@ -3072,12 +3224,56 @@ namespace irods::http::endpoint_operation
...
+						logging::error(
+							*_sess_ptr,
+							"{}: [irods-api-request-resource] and [irods-api-request-replica-number] parameters are "
+						    "incompatible.",
+							__func__);
+						res.result(::http::status::bad_request);
+						res.prepare_payload();
+						return _sess_ptr->send(std::move(res));
+					}
+#endif
```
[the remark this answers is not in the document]

test/test_irods_http_api.py
```diff
@@ -3339,6 +3339,238 @@ def test_rodsadmin_can_invoke_truncate_on_other_users_data_objects_using_the_adm
...
+            with self.subTest('Writing to replica using invalid resource'):
+                r = requests.post(self.url_endpoint, headers=rodsuser_headers, data={
+                    'op': 'write',
+                    'lpath': data_object,
+                    'resource': 'doesnotexist',
+                    'bytes': 'bad resource'
+                })
+                self.logger.debug(r.content)
+                self.assertEqual(r.status_code, 200)
+                self.assertEqual(r.json()['irods_response']['status_code'], irods_error_codes.INVALID_HANDLE)
```
korydraughn, line 3502: Cover different methods of the `write` operation - i.e. header-based form and multipart.

test/test_irods_http_api.py
```diff
@@ -3339,6 +3339,238 @@ def test_rodsadmin_can_invoke_truncate_on_other_users_data_objects_using_the_adm
...
+                    'bytes': 'bad replica number'
+                })
+                self.logger.debug(r.content)
+                self.assertEqual(r.status_code, 400)
+
+            with self.subTest('Writing to replica using incorrect replica number'):
+                r = requests.post(self.url_endpoint, headers=rodsuser_headers, data={
+                    'op': 'write',
+                    'lpath': data_object,
+                    'replica-number': 100,
```
korydraughn, line 3486: Apparently, passing -1 as the replica number results in the iRODS server skipping/ignoring it and continuing as if the user didn't provide it.

See https://github.com/irods/irods/blob/4669cd6be829ed7bc3a4c6648b7eac408839e647/server/core/src/voting.cpp#L81-L84.

I haven't confirmed this with gdb, but as soon as I changed the integer to 100, the server rejected it and the dstream object entered a bad state, as desired.

So, the docs need to be updated to state that the behavior is unspecified if a negative replica number is provided. Or, the HTTP API just disallows negative replica numbers until the iRODS server is updated.

korydraughn (commented):
All tests pass.

test/test_irods_http_api.py
korydraughn: Updated.

trel (commented):
looks good to me.

korydraughn:
The issue numbers have been fixed.

Ready for merging.

alanking (approved):

Issue #271, opened 2024-03-19
Add operation for new `rx_replica_truncate` API


korydraughn:
## Enhancement
See https://github.com/irods/irods/blob/8bf8637b33a27f42029b58a56dc507b1ffc0309d/lib/api/include/irods/replica_truncate.h for details.

PR #455, opened 2025-08-15
[#271] Implement truncate operation for /data-objects endpoint (main)

korydraughn:
Will run the non-OIDC tests today.

As mentioned in #454, the checks for incompatible parameters are disabled. The HTTP API has operations which do these kinds of checks, but I'm leaning towards it shouldn't since that duplicates the work of the iRODS server. One can argue that detecting such situations is good because it protects the iRODS server from needing to do it. One of the goals of the HTTP API is to make interacting with iRODS easier, that means having guard rails around operations.

So yeah, still need to think through if the checks are worth restoring or removing.

### DIFF
--- a/endpoints/data_objects/src/main.cpp
+++ b/endpoints/data_objects/src/main.cpp
@@ -26,6 +26,7 @@
 #include <irods/packStruct.h>
 #include <irods/phyPathReg.h>
 #include <irods/rcMisc.h>
+#include <irods/replica_truncate.h>
 #include <irods/rodsErrorTable.h>
 #include <irods/rodsKeyWdDef.h>
 #include <irods/rodsPackInstruct.h>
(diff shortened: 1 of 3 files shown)


test/test_irods_http_api.py
```diff
@@ -3110,6 +3110,263 @@ def test_server_does_not_segfault_when_parallel_write_init_is_invoked_with_insuf
...
+                'lpath': data_object,
+                'size': 100,
+                'replica-number': 0,
+                'resource': 'demoResc'
+            })
+            self.logger.debug(r.content)
+            self.assertEqual(r.status_code, 200)
+            self.assertEqual(r.json()['irods_response']['status_code'], irods_error_codes.SYS_INVALID_INPUT_PARAM)
+
+            # TODO(???): Enable once the truncate API honors this case.
```
alanking, line 3283: Yep, we should probably make sure this doesn't happen... I don't think any of our APIs enforce this.

alanking (commented):
Everything looks good to me. Just need an issue for those TODO comments, and to make a decision on whether the client enforces the incompatible params.

endpoints/data_objects/src/main.cpp
```diff
@@ -2305,6 +2308,134 @@ namespace
...
+						logging::error(
+							*_sess_ptr,
+							"{}: [resource], [replica-number], and [resource-hierarchy] parameters are incompatible.",
+							fn);
+						return _sess_ptr->send(irods::http::fail(res, http::status::bad_request));
+					}
+#endif
+
+					addKeyVal(&input.condInput, RESC_HIER_STR_KW, iter->second.c_str());
+				}
```
korydraughn, line 2397: After some discussion, we've decided to not expose the resource hierarchy option.

korydraughn:
Addressed all review comments and added documentation.

All non-OIDC tests pass.

alanking (commented):
Looks like clang-format is upset. Otherwise, looks good to me.

korydraughn:
Yep. Squashing and formatting.

korydraughn:
Squashed and formatted.

korydraughn:
Added pound.

PR #461, opened 2025-08-20
[#399] parallel_write_init: Allow targeting replicas via replica number (main)

korydraughn:
All tests pass.

### DIFF
--- a/API.md
+++ b/API.md
@@ -927,7 +927,7 @@ curl http://localhost:<port>/irods-http-api/<version>/data-objects \
     [-F,--data-urlencode] 'op=write' \
     [-F,--data-urlencode] 'lpath=<string>' \ # Absolute logical path to a data object.
     [-F,--data-urlencode] 'resource=<string>' \ # The root resource holding the replica to write to. Optional.
-    [-F,--data-urlencode] 'replica-number=<integer>' \ # The replica number of the replica to write to. Cannot be negative if specified. Optional.
+    [-F,--data-urlencode] 'replica-number=<integer>' \ # The replica number of the replica to write to. Should not be negative if specified. Optional.
     [-F,--data-urlencode] 'offset=<integer>' \ # Number of bytes to skip. Defaults to 0. Optional.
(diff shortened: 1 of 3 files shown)


test/test_irods_http_api.py
```diff
@@ -2412,6 +2412,203 @@ def test_parallel_write_init_operation_returns_an_error_when_targeting_child_res
...
+            self.logger.debug(r.content)
+
+            # Remove the unixfilesystem resource.
+            r = requests.post(f'{self.url_base}/resources', headers=rodsadmin_headers, data={
+                'op': 'remove',
+                'name': ufs_resc
+            })
+            self.logger.debug(r.content)
+
+    def test_parallel_write_init_operation_returns_an_error_on_invalid_replica_number(self):
```
alanking, line 2572: Do we need tests for invalid inputs such as those added in https://github.com/irods/irods_client_http_api/commit/3b8a5b2cd79f7667dac7611b84e198fde3412252 (i.e. `test_targeting_replica_using_invalid_replica_number_or_resource_results_in_an_error`), or do those cover this case as well?

API.md
```diff
@@ -1008,12 +1008,15 @@ curl http://localhost:<port>/irods-http-api/<version>/data-objects \
     --data-urlencode 'op=parallel_write_init' \
     --data-urlencode 'lpath=<string>' \ # Absolute logical path to a data object.
     --data-urlencode 'resource=<string>' \ # The root resource to write to. Optional.
+    --data-urlencode 'replica-number=<integer>' \ # The replica number of the replica to write to. Cannot be negative if specified. Optional.
```
alanking, line 1011: So... the client doesn't actually prevent providing a negative value (from what I can tell). Is the "Cannot" here more of a "Should not" due to https://github.com/irods/irods/issues/8617?

alanking (commented):
Looks good. Couple of questions

test/test_irods_http_api.py
korydraughn: Done.

API.md

korydraughn:
All tests pass.

korydraughn:
Squashed.

korydraughn:
Added pounds.

Issue #477, opened 2026-03-19
HTTP API dies when invalid rule syntax is submitted


trel:
> 
> Sent rule code """hh""" (generally any imperfection with the syntax will kill it)
> 
> ```
> rule_code = """hh"""
> 
> response = client.rules.execute(
>     rule_text=rule_code,
>     rep_instance='irods_rule_engine_plugin-irods_rule_language-instance'
> )
> ```
> 
> which will return 
> 
> ```
> {'status_code': 200, 'data': {'irods_response': {'status_code': -1102000}, 'stderr': None, 'stdout': None}}
> ```
> 
> logs from docker:
> 
> ```
> [2026-02-14 11:56:42.547] [P:1] [info] [T:13] [10.30.1.11] op_list_rule_engines: client_info.username = [test_project_owner]
> double free or corruption (out)
> ```
> 
> After that, the iRODS HTTP API died instantly.
> 

 _Originally posted by @SirValecekLuis in [#471](https://github.com/irods/irods_client_http_api/issues/471#issuecomment-3901876937)_

korydraughn:
I'm just now realizing that this issue is a duplicate of #313.

See PR #486 for fix.

Issue #313, opened 2024-07-25
Execute rule operation, when given non-existing function for 'rule-text' parameter, is causing invalid free to occur when http api is shut down


farbs03:
### iRODS Version, OS and Version

Tip of main

### What did you try to do?

Execute a rule operation with a non-existing function, "writeline" (which was a typo, supposed to be "writeLine") then stop the http api container directly after.

### Expected behavior

`op_execute: RcComm::rError info = [status=[-1220000], message=[DEBUG:]]` gets logged by the http api, and no message indicating an invalid free is occuring gets logged when shutting down the container.

### Observed behavior (including steps to reproduce, if applicable)

The same message above gets logged, except `free(): invalid pointer` gets logged when shutting down the container.

Steps to reproduce:

1. Send execute rule request to the http api with a non-existing function passed to the 'rule-text' parameter
2. Shut down the container running the http api

korydraughn:
I've tried to reproduce this against the following _(all linux-based solutions)_:
- 0.3.0 release published on Docker hub
- A build later than tip-of-main (7b5ed8a)
- A Docker container running a build later than tip-of-main (7b5ed8a)

No attempt produced the results reported.

Given that this issue was reported from use of the Docker for Windows, it's possible the behavior stems from Docker running in a Windows environment.

I'll double check the implementation and see if anything stands out. Aside from that, I don't think there's anything actionable here without being able to reproduce it locally.

PR #486, opened 2026-06-02
[#313,#477] Set rError pointer to nullptr after deallocation (main)

korydraughn:
The test traps a segfault and the fix resolves it.

Using gdb to examine the server's behavior, I see that `freeRError()` is invoked for each RcComm's rError object. That happens at startup, so it's unlikely to be related. It's most likely the connection pool initialization logic. I'll confirm to be on the safe side.

This is ready for review.


korydraughn:
Confirmed the calls to `freeRError` are due to the connection pool init logic. So, no issue there.

korydraughn:
All core tests pass.

trel:
very nice.

test/test_irods_http_api.py
```diff
@@ -4661,6 +4661,30 @@ def test_server_reports_error_when_http_method_is_not_supported(self):
     def test_server_reports_error_when_op_is_not_supported(self):
         do_test_server_reports_error_when_op_is_not_supported(self)
 
+    def test_server_does_not_terminate_on_invalid_rule_syntax(self):
```
trel, line 4664: not sure of the convention of adding the issue number to the test name in this repo...

korydraughn: It was decided a while back to not include issue numbers since the project is still pre-1.0.

I don't feel it's necessary at this time, but I'm not opposed to it.

trel: ah, missed this.  no worries.   resolving.

Issue #470, opened 2026-02-04
Add functionality to change default application name - spOption


mstfdkmn:
As outlined in https://github.com/irods/irods_client_library_rirods/issues/58, clients that depend on irods-http-api, such as rirods, should be able to update their default application name.

PR #488, opened 2026-06-03
[#470] Allow administrators to change ips display name (main)

korydraughn:
- Confirmed `ips` reflects the name defined by the new optional `name` config property
- Confirmed that if `name` is not defined, the HTTP API uses the default name and `ips` reports that name
- Empty strings and strings containing only whitespace are disallowed

### DIFF
--- a/core/src/main.cpp
+++ b/core/src/main.cpp
@@ -343,6 +343,11 @@ constexpr auto default_jsonschema() -> std::string_view
         "irods_client": {
             "type": "object",
             "properties": {
+                "name": {
+                    "type": "string",
+                    "pattern": "[^\\s]",
+                    "maxLength": 255
+                },
                 "host": {
                     "type": "string"
                 },
@@ -525,6 +530,8 @@ auto print_configuration_template() -> void
     }},
 
     "irods_client": {{
+        "name": "irods_http_api",
+
         "host": "<string>",
         "port": 1247,
         "zone": "<string>",
@@ -658,6 +665,12 @@ auto is_valid_configuration(const std::string& _schema_path, const std::string& 
 
 	return false;
 } // is_valid_configuration
+
+auto set_ips_display_name(const json& _config) -> void
+{
+	const auto name = _config.value(json::json_pointer{"/irods_client/name"}, "irods_http_api");
+	set_ips_display_name(name.c_str());
+} // set_ips_display_name
 
 auto set_log_level(const json& _config) -> void
 {
@@ -952,8 +965,6 @@ auto main(int _argc, char* _argv[]) -> int
 	po::positional_options_description pod;
 	pod.add("config-file", 1);
 
-	set_ips_display_name("irods_http_api");
-
 	try {
 		po::variables_map vm;
 		po::store(po::command_line_parser(_argc, _argv).options(opts_desc).positional(pod).run(), vm);
@@ -1000,6 +1011,8 @@ auto main(int _argc, char* _argv[]) -> int
 		}();
 
 		irods::http::globals::set_configuration(config);
+
+		set_ips_display_name(config);
 
 		{
 			const auto schema_file = (vm.count("jsonschema-file") > 0) ? vm["jsonschema-file"].as<std::string>() : "";
--- a/README.md
+++ b/README.md
@@ -345,6 +345,10 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g. 
 
     // Defines iRODS connection information.
     "irods_client": {
+        // The string which is sent to the iRODS server to help identify connections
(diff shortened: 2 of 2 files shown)


korydraughn:
All core tests pass.

core/src/main.cpp
```diff
@@ -343,6 +343,10 @@ constexpr auto default_jsonschema() -> std::string_view
         "irods_client": {
             "type": "object",
             "properties": {
+                "name": {
+                    "type": "string",
+                    "pattern": "[^\\s]"
```
trel: do we need a length min/max?

korydraughn: We can. I'm unsure of what those limits would be.

Do you have a min and max in mind?

korydraughn: To answer your question directly, they aren't required. Applying a max is probably more important than a min.

trel:
is there a max in the spOption field itself on the server side?

core/src/main.cpp
trel: spOption itself?

korydraughn:
There's no max on spOption itself since it's passed in the StartupPack. However, the string is copied into the `RsComm::option` member which has a max buffer size of `LONG_NAME_LEN` (256).

trel:
256 sounds like a good max then.

korydraughn:
`ips` is not capable of showing a name of that length, today. It's truncating the string.

```
Server: localhost                                                                                             
   745517 rods#tempZone  0:00:02  Axxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  IP-REDACTED
   745532 rods#tempZone  0:00:02  Axxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  IP-REDACTED
   745531 rods#tempZone  0:00:02  Axxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  IP-REDACTED
   745594 kory#tempZone  0:00:00  ips  127.0.0.1                                                              
   745533 rods#tempZone  0:00:02  Axxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  IP-REDACTED
   745534 rods#tempZone  0:00:02  Axxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  IP-REDACTED
   745530 rods#tempZone  0:00:02  Axxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  IP-REDACTED
   745529 rods#tempZone  0:00:02  Axxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  IP-REDACTED
   745528 rods#tempZone  0:00:02  Axxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  IP-REDACTED
```

`ips` truncates the string to 63 bytes.

korydraughn:
I don't know what causes that. The file which stores that info contains the truncated string too. The string might be truncated before leaving the client.

Not a big deal.

core/src/main.cpp
korydraughn: See conversation at https://github.com/irods/irods_client_http_api/pull/488#issuecomment-4615290245.

Resolving.

korydraughn:
Max length constraint has been added and works as intended.

Forgot to update the README, so doing that now.

alanking (commented):
Squash em

korydraughn:
Squashed.

core/src/main.cpp
```diff
@@ -343,6 +343,11 @@ constexpr auto default_jsonschema() -> std::string_view
         "irods_client": {
             "type": "object",
             "properties": {
+                "name": {
+                    "type": "string",
+                    "pattern": "[^\\s]",
+                    "maxLength": 255
```
trel, line 349: what happens when someone sets this but they set it to an empty string?

korydraughn: The server rejects empty strings and strings containing only whitespace.

trel: oh!  well there you go.

trel:
i vote to pound it if i get a vote.

korydraughn:
Core tests pass.

Adding pound and merging.

Issue #489, opened 2026-06-03
`parallel_write_init` creates N+1 streams


korydraughn:
If a user specifies a `stream-count` of _N_, `parallel_write_init` should do the following.
- Reserve memory for _N_ streams
- Open the primary stream
- Open _N_-1 secondary streams

`parallel_write_init` does the following, today.
- Reserve memory for _N_ streams
- Open the primary stream
- Open _N_ secondary streams

Issue #482, opened 2026-05-12
`parallel_write_init`: Require `stream-count` to be greater than or equal to 1


korydraughn:
Passing 0 for the stream count doesn't appear to have a negative effect on the HTTP API's behavior. We need to update API.md to include wording about 0 being valid.

The following code snippets are from `parallel_write_init`. They show where the stream count is used.

https://github.com/irods/irods_client_http_api/blob/486c1b120bfc39ea405dce56ac2f09174e6536f5/endpoints/data_objects/src/main.cpp#L1336-L1355

https://github.com/irods/irods_client_http_api/blob/486c1b120bfc39ea405dce56ac2f09174e6536f5/endpoints/data_objects/src/main.cpp#L1362

https://github.com/irods/irods_client_http_api/blob/486c1b120bfc39ea405dce56ac2f09174e6536f5/endpoints/data_objects/src/main.cpp#L1411-L1422

korydraughn:
Marking this as a bug since passing a stream-count of 0 makes no sense.

I'm pretty sure I've identified a bug in `parallel_write_init` too. It creates one too many streams.

trel:
agreed - 0 makes no sense.

PR #490, opened 2026-06-03
[#482,#489] parallel_write_init: Fix handling of stream-count (main)

korydraughn:
- The new test fails when the fix isn't included
- The new test passes when the fix is included
- All core tests pass

### DIFF
--- a/endpoints/data_objects/src/main.cpp
+++ b/endpoints/data_objects/src/main.cpp
@@ -1345,6 +1345,17 @@ namespace
 				}
 
 				const auto stream_count = std::stoi(stream_count_iter->second);
+				if (stream_count < 1) {
+					logging::error(
+						*_sess_ptr,
+						"{}: Argument for [stream-count] parameter is less than 1.",
+						fn,
+						stream_count_iter->second);
+					res.result(http::status::bad_request);
+					res.prepare_payload();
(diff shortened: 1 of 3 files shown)


endpoints/data_objects/src/main.cpp
```diff
@@ -1415,7 +1426,7 @@ namespace
 
 					// Open secondary streams using the first stream as a base.
```
trel: consider adding another line of comment here to explain the math a bit more?

could help the next reader...

korydraughn: It's not clear to me what to talk about. Can you say more about where readers might need more context?

trel: [This issue](https://github.com/irods/irods_client_http_api/issues/489) has the conversation about 1 and then N-1.   A comment in the code saying the same thing would explain the use of starting the counter at 1 in the loop.

```suggestion
					// Open secondary streams using the first stream as a base.
					// Begin this loop at i=1 to produce the remaining n-1 secondary streams
```

korydraughn: Ah, I see now.

Will do.

trel (approved):

Issue #466, opened 2025-10-10
Not working offset and count parameters while using genquery2


MarekNieslanik:
When I run in python

```
query = "SELECT COLL_NAME WHERE COLL_PARENT_NAME LIKE '/tempZone/home/rods%'"
params = {
    'op': 'execute_genquery',
    'query': query,
    'offset': 0,
    'parser': 'genquery1',
    'count': 1000
}
res = requests.get(url + '/query', headers=headers, params=params).json()['rows']
print(len(res))
```
The result contains 257 rows. (I have 257 subcollections in home directory).

When I try to run it with `genquery2`, the result contains only 256 rows. The count parameter doesn't work. It seems that the `count` value is clamped to 256 on the server side. Setting `offset` parameter has no influence, result is the same. In addition I tried to change `max_number_of_rows_per_catalog_query` in the config, but it didn't help.

So when using `genquery1`, both parameters work as expected - count changes the number of rows in the response and offset shifts the results. When switching to `genquery2` I'm not able to get more than 256 rows in the result as the parameters `size` and `offset` don't work.

iRODS Server - 5.0.1
HTTP API - 0.6.0

PR #493, opened 2026-06-04
[#466] Document that offset and count are only supported by GenQuery1 (main)

korydraughn:
The trailing comments/docs on each parameter are good enough for now, but I feel a slight refresh happening in the future.


alanking (approved):
As advertised.

korydraughn:
Added pound.

Issue #494, opened 2026-06-05
Rename `new-data-expiry` to `new-data-expiry-time` to match sibling time parameters


korydraughn:
This is about consistency for the `modify_replica` operation of the /data-objects.

No changes in behavior.

trel:
any need to keep the old name for a release or two?   or 'pre-1.0', so nope.

korydraughn:
Right. No need since it's still pre-1.0.

The release notes will highlight the change.

trel:
sold.

Issue #474, opened 2026-02-22
input time values should be consistent


trel:
These two inputs to data_objects/modify_replica are marked as integers.

https://github.com/irods/irods_client_http_api/blob/486c1b120bfc39ea405dce56ac2f09174e6536f5/API.md?plain=1#L1263-L1264

I think they should be strings.  Same as 

https://github.com/irods/irods_client_http_api/blob/486c1b120bfc39ea405dce56ac2f09174e6536f5/API.md?plain=1#L1266

and
https://github.com/irods/irods_client_http_api/blob/486c1b120bfc39ea405dce56ac2f09174e6536f5/API.md?plain=1#L1275

This *could* just be a documentation exercise.

PR #495, opened 2026-06-05
[#474,#494] Make input placeholder consistent and rename expiry parameter (main)

korydraughn:
All non-OIDC tests pass.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1260,8 +1260,8 @@ curl http://localhost:<port>/irods-http-api/<version>/data-objects \
     --data-urlencode 'replica-number=<integer>' \
     --data-urlencode 'new-data-checksum=<string>' \
     --data-urlencode 'new-data-comments=<string>' \
-    --data-urlencode 'new-data-create-time=<integer>' \
-    --data-urlencode 'new-data-expiry=<integer>' \
+    --data-urlencode 'new-data-create-time=<string>' \
+    --data-urlencode 'new-data-expiry=<string>' \
     --data-urlencode 'new-data-mode=<string>' \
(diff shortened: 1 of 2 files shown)


API.md
```diff
@@ -1260,8 +1260,8 @@ curl http://localhost:<port>/irods-http-api/<version>/data-objects \
     --data-urlencode 'replica-number=<integer>' \
     --data-urlencode 'new-data-checksum=<string>' \
     --data-urlencode 'new-data-comments=<string>' \
-    --data-urlencode 'new-data-create-time=<integer>' \
-    --data-urlencode 'new-data-expiry=<integer>' \
+    --data-urlencode 'new-data-create-time=<string>' \
+    --data-urlencode 'new-data-expiry-time=<string>' \
```
alanking, line 1264: Is it okay to change the interface? I guess we are pre-1.0, so any release is pretty much a major release.

alanking (commented):
I'm guessing that the reason there are no test updates is because we don't really use the expiry field? It's fine not to have a test for it (because we don't use it) but just wanted to check.

API.md
alanking: See https://github.com/irods/irods_client_http_api/issues/494#issuecomment-4634050193. Resolving.

korydraughn:
> I'm guessing that the reason there are no test updates is because we don't really use the expiry field? It's fine not to have a test for it (because we don't use it) but just wanted to check.

Correct. I don't know that anything in the iRODS ecosystem uses the expiry time.

alanking (approved):
\# it

korydraughn:
Added pounds.

Issue #440, opened 2025-07-27
Add version of iRODS server to /info response


korydraughn:
There are times where the tests need to know what version of iRODS the HTTP API is communicating with. For example, modifying the access time of a replica requires the user to know they are communicating with an iRODS 5 server.

The /info endpoint is a good place to expose that information. The question is whether that information should be publicly available.

Here are two ways in which that information can be exposed, if it's considered private/sensitive.
- Require the caller to be a rodsadmin
- Require the caller to provide a passphrase

korydraughn:
That's the version of the HTTP API. This issue is about exposing the version of the iRODS server.

trel:
Oh!  Reading is important.

Okay, yes, the /info should probably give more information if an authenticated user hits it.

Maybe there are different sets for rodsadmin vs rodsuser vs not-authenticated.

korydraughn:
That's possible. The thing I'm thinking about is if the iRODS server has a vulnerability, will reporting the version of the iRODS server be considered sensitive info?

trel:
Right.  Which is why it probably shouldn’t be in the unauthenticated response.   But perhaps in an authenticated response…

PR #499, opened 2026-06-11
[#440] Include iRODS server version in response from /info (main)

korydraughn:
All non-OIDC tests pass.

### DIFF
--- a/core/src/main.cpp
+++ b/core/src/main.cpp
@@ -7,6 +7,7 @@
 #include "irods/private/http_api/process_stash.hpp"
 #include "irods/private/http_api/version.hpp"
 
+#include <irods/client_connection.hpp>
 #include <irods/connection_pool.hpp>
 #include <irods/fully_qualified_username.hpp>
 #include <irods/irods_configuration_keywords.hpp>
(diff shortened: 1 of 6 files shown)


core/src/main.cpp
```diff
@@ -1075,10 +1076,33 @@ auto main(int _argc, char* _argv[]) -> int
...
+			const auto& zone = config.at(json::json_pointer{"/irods_client/zone"}).get_ref<const std::string&>();
+			const auto& username = config.at(json::json_pointer{"/irods_client/proxy_admin_account/username"})
+			                           .get_ref<const std::string&>();
+			irods::experimental::client_connection conn{
+				irods::experimental::defer_authentication,
+				host,
+				port,
+				irods::experimental::fully_qualified_username{username, zone}};
+
+			// Cache the version of the connected iRODS server. This is for endpoints like /info.
```
alanking: It COULD be worth noting here that, even though we fetch the iRODS server version here without authenticating with iRODS, we only return the iRODS server version in the /info endpoint to authenticated HTTP API requests. My first reading of this left me a little confused until I scrolled down to the /info implementation. Up to you.

korydraughn: I was wondering if it was worth being more specific as to why we cache the string. Your comment cleared that right up.

Will add more words.

korydraughn:
Clang-Tidy report should be good too.

All non-OIDC tests pass.

alanking (commented):
Seems fine. Squash

korydraughn:
The commit message is incorrect. Will adjust.

korydraughn:
Updated commit message.

Awaiting approval.

trel (approved):

korydraughn:
Merging.

Issue #484, opened 2026-05-15
`irods_client_http_api` crashes with SIGKILL under load during concurrent `parallel_write_init` requests


mstfdkmn:
HTTP API Version:0.6.0
iRODS Server:5.0.2 

In one of our use cases, we found that the API crashes when too many concurrent connections are made using a single account. I reproduced this issue. When I send 40 concurrent parallel_write_init requests to the HTTP API, some succeed but most crash the server.
The daemon hits a memory fault during cleanup (ssl_client_stop / rcDisconnect) and the OS kills it with SIGKILL.
This causes a cascade of 502 errors from the proxy.

Expected Behavior: Error responses, not a dead server. We think the API should never crash. Worst-case, the requests should be rejected, queued, or ignored but it should not lead to an unresponsive service.
Observed Behavior: Server dies. 502 Bad Gateway errors

Steps to reproduce:
POST to /irods-http-api/0.6.0/data-objects
Use op=parallel_write_init, stream-count=2, unique lpath each time

I used the following python script to reproduce:
```
import asyncio
import aiohttp
import os
import time

URL = "https://icts.irods.t.icts.kuleuven.be/irods-http-api/0.6.0/data-objects"
TOKEN = token
BASE_PATH = "/icts/home/u0137480/test_parallel.bin"

async def send_request(session, idx, headers):
    data = aiohttp.FormData(quote_fields=False)
    data.add_field("op", "parallel_write_init")
    data.add_field("lpath", f"{BASE_PATH}_{idx}_{int(time.time())}")
    data.add_field("stream-count", "2")
    try:
        async with session.post(URL, data=data, headers=headers) as resp:
            return resp.status
    except:
        return "Network Failure"

async def main():
    if not TOKEN:
        print("No ACCESS_TOKEN set")
        return

    headers = {"Authorization": f"Bearer {TOKEN}", "Accept": "application/json"}
    connector = aiohttp.TCPConnector(limit=None, force_close=True)

    async with aiohttp.ClientSession(connector=connector) as session:
        tasks = [send_request(session, i, headers) for i in range(40)] # FYI: during my test changing this number to 33 worked normal
        results = await asyncio.gather(*tasks)

        counts = {r: results.count(r) for r in set(results)}
        for status, n in counts.items():
            print(f"{status}: {n}")

if __name__ == "__main__":
    asyncio.run(main())
```

Log Traces:
```
[...]

Issue #79, opened 2023-08-27
Add configurable limit on the number of parallel streams that can be active/instantiated at one time


korydraughn:
The HTTP API currently allows an infinite number of parallel streams to be opened. This is really bad for systems which cannot afford to go down often.

To help with this situation, the HTTP API needs to provide a config option to limit the number of active parallel streams. This protects the HTTP API and the iRODS server.

If a client requests _N_ streams and instantiating them results in exceeding the limit, the HTTP API needs to return an appropriate HTTP status code to inform the client that they should try again later due to resource limitations.

trel:
This is independent of the current settings for number of threads incoming, and size of connection pool to iRODS...

So this is a total, across all threads/connections.   Will have to do some min/max comparisons on the configs to make sure they make sense together.

korydraughn:
Correct. This has more to do with how parallel writes are supported.

When a client wants to do a parallel write, they have to tell the HTTP API how many streams they want. We already have an option for that limiting that. What we don't have is an option to limit the total number of streams across all clients.

Initiating parallel writes causes the HTTP API to create dedicated dstream objects for a client. They are not shared and they persist across requests.

trel:
But we already have 

```
        "requests": {
            // The number of threads dedicated to servicing client requests.
            // When adjusting this value, consider adjusting "background_io/threads"
            // and "irods_client/connection_pool/size" as well.
            "threads": 3,
```

Does that not limit the total itself?

korydraughn:
It doesn't.

That only limits the number of requests that can be **accepted** asynchronously.

The HTTP API can create any number of dstream objects. It does not consider anything other than the parallel write stream count limit when dealing with parallel writes.

trel:
Okay, so we'll have 3-4-5 different limits that will interact with each other, kind of...

korydraughn:
Correct.

PR #501, opened 2026-06-15
[#79,#484] Limit parallel-write streams across all clients (main)

korydraughn:
All tests pass.


test/test_irods_http_api.py
```diff
@@ -4037,6 +4037,98 @@ def test_stat_does_not_expand_group_permissions(self):
                 })
                 self.logger.debug(r.content)
 
+    def test_server_returns_http_status_code_503_when_active_parallel_write_stream_count_exceeds_limit(self):
+        # This test assumes the HTTP API is configured to allow no more than 15
+        # parallel-write streams in the system and that no more than 3 streams
+        # can be created per parallel_write_init invocation.
+        #
+        # TODO(?): Update this test to automatically skip/fail based on the
+        # HTTP API's configuration.
```
korydraughn: Add issue number to TODO.

README.md
```diff
@@ -460,9 +460,14 @@ Notice how some of the configuration values are wrapped in angle brackets (e.g.
...
 
+        // The maximum number of parallel streams that can be reserved by
+        // the HTTP API across all clients. Exceeding this limit will result
+        // in an HTTP status code of 503 (Service Unavailable).
+        "max_number_of_active_parallel_write_streams": 15,
+
         // The maximum number of parallel streams that can be associated to a
         // single parallel write handle.
-        "max_number_of_parallel_write_streams": 3,
+        "max_number_of_streams_per_parallel_write_handle": 3,
```
alanking, line 469: So, `max_number_of_streams_per_parallel_write_handle` is the "per-client" max number of parallel write streams, and `max_number_of_active_parallel_write_streams` is the overall max number of parallel write streams that are allowed in an HTTP API server.

Where did the default value of 15 for `max_number_of_active_parallel_write_streams` come from?

endpoints/data_objects/src/main.cpp
```diff
@@ -1346,28 +1351,45 @@ namespace
...
+							*_sess_ptr,
+							"{}: Argument for [stream-count] parameter would exceed maximum number of parallel streams "
+							"allowed by system: stream-count=[{}], active=[{}]",
+							fn,
+							stream_count,
+							active);
+						res.result(http::status::service_unavailable);
+						res.prepare_payload();
+						return _sess_ptr->send(std::move(res));
+					}
```
alanking, line 1404: It is a subtle, perhaps inconsequential thing, but we would enter this codepath if `max_number_of_streams_per_parallel_write_handle` > `max_number_of_active_parallel_write_streams`. Should we consider that a configuration error?

test/test_irods_http_api.py
```diff
@@ -4037,6 +4037,98 @@ def test_stat_does_not_expand_group_permissions(self):
...
+                self.assertGreater(len(result['parallel_write_handle']), 0)
+                parallel_write_handles.append(result['parallel_write_handle'])
+
+            # Show that exceeding the active stream limit is detected and results
+            # in the expected HTTP status code.
+            other_data_object = data_object + '.nope'
+            r = requests.post(self.url_endpoint, headers=rodsuser_headers, data={
+                'op': 'parallel_write_init',
+                'lpath': other_data_object,
+                'stream-count': 2
```
alanking, line 4099: Just curious: Why is the `stream-count` different for each `parallel_write_init` invocation? It is 3 for the loop above on R4063 and 1 for the other-user invocation below on R4098.

test/test_irods_http_api.py
```diff
@@ -4037,6 +4037,98 @@ def test_stat_does_not_expand_group_permissions(self):
...
+        try:
+            data_object = f'/{self.zone_name}/home/{self.rodsuser_username}/issues_79_and_484.txt'
+
+            # Invoke parallel_write_init until 15 streams exist in the system.
+            for i in range(5):
+                parallel_write_data_objects.append(data_object + f'.{i}')
+                r = requests.post(self.url_endpoint, headers=rodsuser_headers, data={
+                    'op': 'parallel_write_init',
+                    'lpath': parallel_write_data_objects[-1],
+                    'stream-count': 3
```
alanking, line 4084: Should we also have a test in which `stream-count` > 3?

alanking (commented):
Seems reasonable

README.md
korydraughn: Just a random number I pulled out the sky. I expect admins to change it for their deployment.

endpoints/data_objects/src/main.cpp
korydraughn: I think it's safe to treat it as a config error.

test/test_irods_http_api.py
korydraughn: I was trying to show that the http api doesn't treat the values special - i.e. catch bugs.

test/test_irods_http_api.py
korydraughn: I'm surprised there's no test for that already.

Will add one.

endpoints/data_objects/src/main.cpp
korydraughn: I'll add more words around that in the config notes. Will open an issue to add more checks to server startup.

README.md
korydraughn: Rereading your question ...

> So, `max_number_of_streams_per_parallel_write_handle` is the "per-client" max number of parallel write streams ...

No. It is the per-handle (parallel-write handle) limit. A client is allowed to open multiple parallel write streams.

alanking: Okay, so there are three numbers:

A. `max_number_of_streams_per_parallel_write_handle`
B. `max_number_of_parallel_write_handles_per_client` (does not exist, but has a practical limit based on C)
C. `max_number_of_active_parallel_write_streams`

Just trying to make sure I'm following

korydraughn: That's correct.

alanking: Okay, questions answered. Resolving

test/test_irods_http_api.py

endpoints/data_objects/src/main.cpp
korydraughn: I added a post-jsonschema-validation step to startup to disallow that situation.

Works as intended.

test/test_irods_http_api.py
korydraughn: Made both cases use `2`. Using different values just leads to confusion.

test/test_irods_http_api.py
korydraughn: Done.

korydraughn:
Rebased to resolve conflict.

korydraughn:
All non-OIDC tests pass.

core/src/main.cpp
```diff
@@ -671,7 +671,29 @@ auto is_valid_configuration(const std::string& _schema_path, const std::string&
 	}
 
 	return false;
-} // is_valid_configuration
+} // config_meets_jsonschema_requirements
+
+auto config_meets_post_jsonschema_validation_requirements(const json& _config) -> bool
```
alanking, line 676: This is checking things that cannot be enforced in the schema, but I don't think it is important that the validation is happening after we've checked the schema requirements. Thoughts on this name, or something similar?

`config_meets_non_jsonschema_validation_requirements`

alanking (commented):
Just noting that we didn't just rename an existing configuration, but also added a new one which has the old name of the existing configuration. Will need to put that in big letters for the release notes

korydraughn:
> Just noting that we didn't just rename an existing configuration, but also added a new one which has the old name of the existing configuration.

Correct. The inclusion of `_active_` in the config property's name didn't add any value. We are redefining what an existing property means though.

> Will need to put that in big letters for the release notes

Agreed.

core/src/main.cpp
korydraughn: Well, it does depend on schema validation succeeding.

The new function leans on that to avoid dealing with type conversion errors. It assumes the configuration matches the type expectations of the HTTP API.

If the new name is preferred, I'd only need to add a comment to make sure the next person understands the situation.

alanking: That's fair. I guess we can leave it - it's just a helper function, in the end

alanking (commented):
Okay, squash to taste

korydraughn:
Squashed. Please eyeball commit messages for typos, grammar.

alanking (commented):
Just had one suggestion on wording...

On "Limit parallel-write streams across all clients":
"Exceeding this limit..." might sound better than "Violating this limit..."

korydraughn:
Updated commit message.

alanking (approved):

alanking:
Oops, # it

korydraughn:
Added pounds.

Issue #479, opened 2026-04-14
Error propagation failure when violating Logical Quotas via iRODS HTTP API


MarekNieslanik:
Environment:
    iRODS: 5.0.2
    iRODS HTTP API: 0.6.0
    Plugin: irods_rule_engine_plugin_logical_quotas

We are integrating the Logical Quotas Rule Engine Plugin into our iRODS stack to track and limit collection capacity. While the plugin correctly enforces quotas at the server level (for example with icommands), the resulting error messages are not being properly propagated through the iRODS HTTP API.

When a quota constraint is violated during a write operation via the HTTP API, the API returns a generic error that does not reflect the actual cause (the quota violation).

**parallel_write_init failure:**
The HTTP API returns a generic INVALID_HANDLE error:
JSON response from the server
```
{"irods_response":{"status_code":-175000,"status_message":"INVALID_HANDLE: Could not open output stream to [/tempZone/home/rods/abcdef].\n\n"}}  // similar in HTTP API logs
```
However, the iRODS server logs correctly identify the issue:
```
"log_message":"Logical Quotas Policy Violation: Adding object exceeds maximum data size in bytes limit"
```

**Simple write failure**
A similar issue occurs with the standard `write` operation. The error message changes to `Output stream to data object is in a bad state`, but the error code remains -175000.

The main issue is that the error code (`-175000`) and the error messages are too generic. They are identical to errors triggered by other issues, such as a missing target collection or data object.

As a result, it is currently impossible for the client application to distinguish between a policy violation (quota exceeded) and a system error (missing path). The specific error information from the Rule Engine Plugin is lost during the transition through the HTTP API.

We would neet the HTTP API to propagate the specific error message or a unique error code provided by the Logical Quotas plugin, allowing clients to identify that a quota violation has occurred.

MarekNieslanik:
Is there possibility that release 5.1.0 of the iRODS server would fix that?

alanking:
> Is there possibility that release 5.1.0 of the iRODS server would fix that?

I believe a change to the HTTP API would be required to get this to work, not the iRODS server. It is using the streaming library which makes failures harder to understand (because it is based on C++ I/O streams which do the same thing: https://cppreference.com/w/cpp/io.html#Stream-based_I.2FO). Although I might be wrong about some/all of that.

korydraughn:
[...]

PR #502, opened 2026-06-16
[#479] Return iRODS error codes for write operations (main)

korydraughn:
All non-OIDC tests pass.

I've verified the following operations report error codes from the Logical Quotas rule engine plugin too. The error codes come from the plugin and iRODS server.
- `write` - returns `SYS_NOT_ALLOWED` on violations
- `parallel_write_init` - returns `SYS_NOT_ALLOWED` on violations
- `touch` - returns `RE_RUNTIME_ERROR`

Testing covered limits on the number of data objects and bytes.

I think this puts the HTTP API in a good spot. Improvements to error codes have to be addressed in the plugin and/or the iRODS server.


korydraughn:
Forgot to add the compile-time logic for when the application is compiled against iRODS 5.0.2 and earlier.

Will do that next.

korydraughn:
Compiled against 4.3 and irods/irods tip-of-main.

Tested each build against irods/irods tip-of-main.

This is ready for review now.

endpoints/data_objects/src/main.cpp
```diff
@@ -152,7 +152,11 @@ namespace
 			}
 
 			if (!stream_) {
+#if IRODS_VERSION_INTEGER >= 5001000
```
alanking: Should we be using a feature test macro here (i.e. `IRODS_LIBRARY_FEATURE_DSTREAM >= 202606L`)?

test/test_irods_http_api.py
```diff
@@ -41,6 +41,18 @@ def setup_class(cls, opts):
...
+    # configuration indicates that dstream exposes iRODS error codes. Otherwise,
+    # the generic error code (i.e. INVALID_HANDLE) is returned.
+    #
+    # The test configuration option should be enabled whenever the HTTP API is
+    # compiled against the iRODS 5.1.0 development library or later.
+    def identity_or_invalid_handle_impl(self, irods_error_code):
+        if config.test_config.get('dstream_exposes_irods_error_codes', False):
+            return irods_error_code
+        return irods_error_codes.INVALID_HANDLE
+    cls.identity_or_invalid_handle = identity_or_invalid_handle_impl
```
alanking: What is the purpose of this declaration rather than just calling the method directly where needed?

API.md
```diff
@@ -990,8 +990,6 @@ If an HTTP status code of 200 is returned, the body of the response will contain
 
 If there was an error, expect an HTTP status code in either the 4XX or 5XX range.
 
-If the data object cannot be opened, the HTTP API will return an iRODS error code of -175000 (`INVALID_HANDLE`). This error code indicates that the data object may be locked.
```
alanking, line 993: Should we leave a version of this in the docs? Something like this:
```suggestion
If the data object cannot be opened, the HTTP API will return an appropriate iRODS error code. This error code indicates that the data object may be locked.
```

endpoints/data_objects/src/main.cpp
korydraughn: Yes, that was one of the reasons for adding that macro. Somehow I completely forgot about it.

Will update.

test/test_irods_http_api.py
korydraughn: The original implementation was very different and relied on state in the class instance. The function no longer depends on state within the class instance, so it can be a free function now.

I missed that due all the refactoring. Will update.

API.md
korydraughn: I don't think necessary now that dstream reports the actual errors. Before this PR, that documentation made sense to have because failed interactions with dstream almost always produced `INVALID_HANDLE`.

The wording you presented is better than what I came up with (before I decided to remove it), but I don't feel it adds much since 99% of the HTTP API interactions are call API, check status.

Happy to be persuaded otherwise.

alanking: Nah, that sounds fine. Let's leave it out. Resolving.

endpoints/data_objects/src/main.cpp

test/test_irods_http_api.py

korydraughn:
Confirmed that tests pass when compiled against iRODS 5.

Working on a build which compiles against a dev package that doesn't have the dstream enhancement.

korydraughn:
Confirmed that a build against the 4.3 dev package passes the tests.

Awaiting feedback.

endpoints/data_objects/src/main.cpp
```diff
@@ -152,11 +153,11 @@ namespace
 			}
 
 			if (!stream_) {
-#if IRODS_VERSION_INTEGER >= 5001000
+#ifdef IRODS_LIBRARY_FEATURE_DSTREAM
```
alanking, line 156: Does checking the value matter? I know that `IRODS_LIBRARY_FEATURE_DSTREAM` was introduced specifically for this and hasn't changed, so this will work.

alanking (commented):
One last question

endpoints/data_objects/src/main.cpp
korydraughn: Not for this case. It doesn't hurt to check it, but it's unnecessary since we know this is a new macro.

alanking: Alright, let's just leave it.

alanking (commented):
Squash

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Added pounds.

